PR #16, opened 2020-10-27
Generics

JonasWanke:

### DIFF
--- a/packages/compiler/lib/src/compilation/ast_hir_lowering/declarations/property.dart
+++ b/packages/compiler/lib/src/compilation/ast_hir_lowering/declarations/property.dart
@@ -9,7 +9,6 @@ import '../body.dart';
 import '../type.dart';
 import 'declarations.dart';
 import 'module.dart';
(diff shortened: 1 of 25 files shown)


packages/compiler/lib/src/compilation/ast_hir_lowering/declarations/property.dart
```diff
@@ -39,15 +38,6 @@ final getPropertyDeclarationHir = Query<DeclarationId, hir.PropertyDeclaration>(
...
     }
-    if (propertyAst.initializer == null && declarationId.parent.isNotTrait) {
-      throw CompilerError.propertyInitializerMissing(
-        'Properties must have a default value, unless they are declared within a trait.',
-        location: ErrorLocation(
-          declarationId.resourceId,
-          propertyAst.representativeSpan,
-        ),
-      );
-    }
```
MarcelGarus: Properties in modules and impls still need default values.

MarcelGarus (approved):

PR #29, opened 2020-11-10
Improve core

MarcelGarus:


packages/candy/src/collections/iterable.candy
```diff
@@ -1,91 +1,294 @@
...
+      counter = counter + 1
+      counter > n
+    })
+  }
+
+  fun skipWhile(checker: (Item) => Bool): Iterable<Item> {
+    mut let skipping = true
+    where({
+      if !skipping {
+        return false
```
JonasWanke:
```suggestion
        return true
```

packages/candy/src/collections/list/array.candy
```diff
@@ -62,21 +63,21 @@ public class ArrayList<Item> {
...
     )
   }
 
-  mut let items: Array<Option<Item>>
+  mut let items: Array<Maybe<Item>>
   mut let length_: Int /* UInt */
 
   // TODO(marcelgarus): Getterize!
-  fun capacity(): Int /* UInt */ { items.length }
+  public fun capacity(): Int /* UInt */ { items.length }
```
JonasWanke:
```suggestion
  fun capacity(): Int /* UInt */ { items.length }
```

packages/candy/src/collections/map/module.candy
```diff
@@ -0,0 +1,38 @@
...
+
+  fun clear() {
+    // TODO(marcelgarus): Implement
+    // for key in keys() {
+    //   remove(key)
+    // }
+    todo("Implement MutableMap.clear")
+  }
+
+  fun addAll(map: Map<Key, Value>)
```
JonasWanke: Maybe `setAll`?

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+    let segments = string.split(fileSystem.separator()).mutable()
+    let isAbsolute = segments.first == Some("")
+    if isAbsolute {
+      segments.removeFirst()
+    }
+
+    let segments = segments.map {
+      // TODO(marcelgarus): Matchify!
+      if it.isEmpty {
+        ChildSegment(isEmpty = true)
```
JonasWanke:
```suggestion
        CurrentSegment(isEmpty = true)
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+trait FileSystem {
+  // TODO(marcelgarus): Getterize!
+  builtin static fun default()
+
+  // TODO(marcelgarus): Give it a different value based on the platform.
+  public let separator = "/"
+  public let currentDirectorySymbol = "."
+  public let parentDirectorySymbol = ".."
+
+  public fun parseToPath(string: String) => Result<Path> {
```
JonasWanke:
```suggestion
  public fun parseToPath(string: String): Result<Path> {
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+        ParentSegment()
+      } else {
+        ChildSegment(it)
+      }
+    }.toList()
+
+    Path.fromSegments(segments, isAbsolute)
+  }
+
+  public fun stringifyPath(path: Path) => String {
```
JonasWanke:
```suggestion
  public fun stringifyPath(path: Path): String {
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+class Path {
+  /// A slice of a path.
+  ///
+  /// This type is independent of the file system (if `/` or `\` is used). It supports a number of
+  /// operations for working with and inspecting a path, including getting the segments, extracting
+  /// the file name, and so on.
+  /// It ensures that the value is valid, e.g. if it's an absolute path, the path can't escape past
+  /// the root directory.
+
+  static fun unsafe(segments: List<Segment>, isAbsolute: Bool) => Path {
```
JonasWanke:
```suggestion
  static fun unsafe(segments: List<Segment>, isAbsolute: Bool): Path {
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+    /// Creates a Path instance without checking if it's valid.
+    todo()
+  }
+
+  public static fun fromSegments(segments: List<Segment>, isAbsolute: Bool) => Result<Path, ...> {
```
JonasWanke:
```suggestion
  public static fun fromSegments(segments: List<Segment>, isAbsolute: Bool): Result<Path, ...> {
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+            return Error("An absolute path can't escape the root directory!")
+          }
+        }
+      }
+    }
+    Path.unsafe(segments, isAbsolute)
+  }
+
+  public static fun fromString(string: String, fileSystem: FileSystem = FileSystem.default())
+    => Result<Path, ...> {
```
JonasWanke:
```suggestion
  public static fun fromString(string: String, fileSystem: FileSystem = FileSystem.default()): Result<Path, ...> {
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+    fileSystem.parsePath(string)
+  }
+
+  static fun join(first: Path, second: Path) => Result<Path, ...> {
```
JonasWanke:
```suggestion
  static fun join(first: Path, second: Path): Result<Path, ...> {
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+    if first.isAbsolute && second.isAbsolute {
+      Error("Can't combine two absolute paths.")
```
JonasWanke: One use case that comes to my mind regarding this function: A CLI takes a path from the user, and if it's relative, it should be interpreted relative to the current/project directory. In that case, an absolute path is perfectly valid, and the current/project directory just isn't necessary for evaluating the path.

```suggestion
      second
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+      )
+    }
+  }
+
+  let segments: List<Segment>
+  let isAbsolute: Bool
+  /// Whether this `Path` is absolute, i.e. if it's independent of the current directory.
+
+  // TODO(marcelgarus): Getterize!
+  fun isRelative(): Bool => !isAbsolute
```
JonasWanke:
```suggestion
  fun isRelative(): Bool { !isAbsolute }
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+      Path.fromSegments(
+        segments = first.segments + second.segments,
+        isAbsolute = first.isAbsolute && second.isAbsolute,
+      )
+    }
+  }
+
+  let segments: List<Segment>
+  let isAbsolute: Bool
+  /// Whether this `Path` is absolute, i.e. if it's independent of the current directory.
```
JonasWanke:
```suggestion
  /// Whether this `Path` is absolute, i.e., if it's independent of the current directory.
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+
+  // TODO(marcelgarus): Getterize!
+  fun isRelative(): Bool => !isAbsolute
+
+  fun normalize(): Path {
```
JonasWanke:
```suggestion
  fun normalized(): Path {
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+    /// Returns a normalized version of this path that points to the same resource.
+
+    let normalized = MutableList<Segment>()
+    for segment in segments {
+      // TODO(marcelgarus): Matchify!
+      if segment is ChildSegment {
+        normalized.add(segment)
+      } else if segment is ParentSegment {
+        normalized.removeLast()
```
JonasWanke: There might not be a (remaining) segment if the path leaves its root directory. In that case, we should append a `ParentSegment` (if this is a relative path).

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+    }
+    Path.unsafe(normalized, isAbsolute)
+  }
+
+  // TODO(marcelgarus): Getterize!
+  fun parent(): Option<Path> {
+    if segments.isEmpty {
+      None()
+    } else {
+      Some(Path.unsafe(segments.withoutLast(), isAbsolute))
```
JonasWanke: It's not that easy – the last segment might be a `ParentSegment` or `CurrentSegment`. We could normalize it first and then match the last segment, either removing it (if it's a `ChildSegment`), appending a `ParentSegment` (if it's a `ParentSegment`, and in case of an absolute path, we don't leave the root directory), or returning none.

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+  // TODO(marcelgarus): Walk up the path.
+  fun ancestors(): Iterable<Path> {
+    todo()
+  }
+
+  fun startsWith(prefix: List<Segment>): Bool {
+    segments.startsWith(prefix)
+  }
+
+  fun commonDirectory(other: Path) {
```
JonasWanke:
```suggestion
  fun commonDirectory(other: Path): Option<Path> {
```

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+    assert(isAbsolute == other.isAbsolute);
```
JonasWanke: We could return `None` in that case

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+    Path.unsafe(
+      segments.zip(other.segments).takeWhile { a, b => a == b }.map { a, _ => a }.toList()
```
JonasWanke: Paths have to be normalized first

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+  }
+  class CurrentSegment {
+    let isEmpty = false
+    /// The current path can be represented as "." (as in "some/./path") and "" (as in "some//path").
+  }
+  class ParentSegment
+}
+
+impl Path: OperatorSlash<Path>, ToString, Iterable<Path.Segment> {
+  fun slash(Path other) => Path.join(segments, other.segments)
```
JonasWanke:
```suggestion
  fun slash(Path other): Path { Path.join(segments, other.segments) }
```

packages/candy/src/io/module.candy
```diff
@@ -0,0 +1 @@
+public use ..print
```
JonasWanke: Shouldn't this be `.print`?

packages/candy/src/io/file.candy
```diff
@@ -0,0 +1,186 @@
...
+  fun doesExist(): Result<Bool, ...>
+  fun read(): Result<String, ...>
+  fun write(content: String): Result<Unit, ...>
+}
+impl File: FileSystemNode
+
+trait Directory {
+  let path: Path
+}
+impl File: FileSystemNode
```
JonasWanke:
```suggestion
impl Directory: FileSystemNode
```

packages/compiler/lib/src/compilation/hir/ids.dart
```diff
@@ -159,7 +159,8 @@ abstract class ModuleId implements _$ModuleId {
       ModuleId(PackageId.core, ['operators', 'logical']);
   static const corePrimitives = ModuleId(PackageId.core, ['primitives']);
   static const coreReflection = ModuleId(PackageId.core, ['reflection']);
-  static const coreStdio = ModuleId(PackageId.core, ['stdio']);
+  static const coreIo = ModuleId(PackageId.core, ['io']);
+  static const corePrint = ModuleId(PackageId.core, ['io', 'print']);
```
JonasWanke: Should be `coreIoPrint` to be consistent with the other properties.

packages/candy/src/primitives/module.candy
```diff
@@ -1,17 +1,50 @@
...
+/// no information.
+///
+/// Here are a few examples of expressions that evaluate to `Unit`:
+///
+/// * assignments like `let a = 5`
+/// * function calls of functions without an explicit return value like `print("Hello, world!")`
+/// * the explicit unit instance: `unit`
+
 public builtin class Never
+/// Can never be created. It's still useful to specify that an expression will never evaluate. More
```
JonasWanke:
```suggestion
/// Can never be created. It's still useful to specify that an expression will never return directly. More
```

packages/compiler/lib/src/compilation/ast_hir_lowering/body.dart
```diff
@@ -391,24 +392,35 @@ abstract class Context {
...
         CompilerError.ambiguousExpression(
           'Expression is ambiguous.',
           location: ErrorLocation(resourceId, expression.span),
+          relatedInformation: [
+            for (final it in lowered)
+              ErrorRelatedInformation(
+                message: 'This is one of the ambigous options: '
+                    '${it.id.declarationId}',
+                location: ErrorLocation(
+                    it.id.declarationId.resourceId, SourceSpan(0, 0)),
```
JonasWanke:
```suggestion
                location: ErrorLocation(it.id.declarationId.resourceId),
```

packages/compiler/lib/src/compilation/ast_hir_lowering/body.dart
```diff
@@ -391,24 +392,35 @@ abstract class Context {
   ) {
     final result = lower(expression);
     if (result is Error) return Error(result.error);
+    final lowered =
+        result.value.distinctBy((it) => it.id.declarationId).toList();
```
JonasWanke: This should be moved to `findIdentifierInUseLines`

playground/src/main.candy
```diff
@@ -0,0 +1,3 @@
+fun main() {
+  let a = List.of3<int>(1, 2, 3);
+}
```
JonasWanke: If this code should stay, please move it to a named file

JonasWanke (changes requested):

packages/candy/src/io/file.candy
```diff
@@ -57,16 +57,16 @@ class Path {
...
     todo()
   }
 
-  public static fun fromSegments(segments: List<Segment>, isAbsolute: Bool) => Result<Path, ...> {
+  public static fun fromSegments(segments: List<Segment>, isAbsolute: Bool): Result<Path, ...> {
     if isAbsolute {
       // Make sure the path doesn't escape the root directory.
       mut let depth = 0
-      for segment in segments {
+    for segment in segments {
```
JonasWanke:
```suggestion
      for segment in segments {
```

packages/candy/src/io/file.candy
```diff
@@ -112,18 +123,30 @@ class Path {
...
+        }
       }
     }
     Path.unsafe(normalized, isAbsolute)
   }
 
   // TODO(marcelgarus): Getterize!
   fun parent(): Option<Path> {
-    if segments.isEmpty {
+    while segments.last() is Option<CurrentSegment> {
```
JonasWanke, line 139: Shouldn't we make a mutable copy of that?

JonasWanke (changes requested):

packages/candy/src/io/file.candy
MarcelGarus: Jup

JonasWanke (approved):

PR #30, opened 2020-11-12
As operator

MarcelGarus:
Merge `improve-core` first and then merge the `master` into this before reviewing this.


packages/candy/src/collections/iterable.candy
```diff
@@ -111,7 +111,7 @@ public trait Iterable<Item> {
 
   fun cast<Type>(): Iterable<Type> {
     // TODO(marcelgarus): Enable as soon as "as" is supported.
-    // map<Type>({ it as Type })
+    map<Type>({ it as Type })
```
JonasWanke, line 113: The comment in the line above is now deprecated

packages/parser/lib/src/parser/grammar.dart
```diff
@@ -600,13 +600,14 @@ class ParserGrammar {
...
-      ..leftComplex<OperatorToken, Expression>(
-        (LexerGrammar.WS &
+      ..postfix<List<dynamic>, Expression>(
+        LexerGrammar.NLs &
                 (LexerGrammar.AS | LexerGrammar.AS_SAFE) &
-                LexerGrammar.WS)
-            .map((it) => it[1] as OperatorToken),
-        mapper: (left, operator, right) =>
-            BinaryExpression(_id++, left, operator, right),
+                LexerGrammar.NLs &
```
JonasWanke: Are you sure that `NLs` works correctly? `WS` requires some whitespace, whereas `NLs` only allows some optional whitespace.

JonasWanke (changes requested):

packages/compiler/lib/src/compilation/ast_hir_lowering/body.dart
```diff
@@ -392,7 +400,8 @@ abstract class Context {
   ) {
     final result = lower(expression);
     if (result is Error) return Error(result.error);
-    final lowered = result.value.toList();
+    final lowered =
+        result.value.distinctBy((it) => it.id.declarationId).toList();
```
JonasWanke: This basically only takes the first element, as `it.id.declarationId` is the declaration ID of the current body. I don't think this is what we want.

JonasWanke (changes requested):

packages/compiler/lib/src/compilation/ast_hir_lowering/body.dart
```diff
@@ -1638,6 +1636,16 @@ extension on Context {
               );
             }
           })
+          // If one method is defined in multiple places, but is actually the
+          // same one (like `next`, which is defined on both `ArrayList` and
+          // `Iterator`), the expression would be ambiguous. So, for now we work
+          // around this by only considering methods ambigious defined in the
```
JonasWanke:
```suggestion
          // around this by only considering methods ambiguous defined in the
```

JonasWanke (approved):

PR #33, opened 2020-11-16
Fix candy core

MarcelGarus:


packages/compiler_dart/lib/src/declarations/trait.dart
```diff
@@ -1,15 +1,18 @@
 import 'package:code_builder/code_builder.dart' as dart;
 import 'package:compiler/compiler.dart';
 import 'package:compiler_dart/src/constants.dart';
+import 'package:compiler_dart/src/utils.dart';
```
JonasWanke: These two imports should be relative.

JonasWanke (approved):

PR #37, opened 2020-11-18
Add Random

JonasWanke:


packages/core/src/random/source.candy
```diff
@@ -0,0 +1,14 @@
...
+  fun generateIntegers(length: Int): List<Int>
+
+  // TODO(JonasWanke): use bytes when we have them
+  // fun generateBytes(length: Int): List<Int>
+}
+
+builtin class DefaultRandomSource {
+  static fun withSeed(seed: Int): This
+}
+builtin impl DefaultRandomSource: RandomSource
```
MarcelGarus: Do we really want to expose `DefaultRandomSource` to the rest of the Candy ecosystem, or should we rather provide static methods on `RandomSource`, as in `RandomSource.default()` and `RandomSource.defaultWithSeed(...)`?

packages/core/src/random/source.candy
```diff
@@ -0,0 +1,14 @@
+use ...collections
+use ...primitives
+
+trait RandomSource {
+  fun generateIntegers(length: Int): List<Int>
+
+  // TODO(JonasWanke): use bytes when we have them
+  // fun generateBytes(length: Int): List<Int>
```
MarcelGarus: I'm not sure about this one. I think we should offer randomness at smaller granularities:
- `generateBool` for generating a bit
- `generateByte` for generating a byte
- `generateBytes` for generating an arbitrary number of bytes

Of course, for now, the byte generators could return `int`s up to 255 instead.

Some advantages:

- We could provide default implementations for everything except `generateBool`. That makes it very easy to implement `RandomSource` while keeping the option to write more efficient `RandomSource`s. ("Very easy" as in "just write a method that returns `true` 50% of the time".)
- Providing randomness on bit granularity also has the advantage that we are more careful about using up entropy. When we implement a secure `RandomSource`, we'll probably use something like /dev/random on Linux and we could use bitshifting to give out single bits of entropy. That seems especially useful for generating bools and enum variants without consuming a whole byte of entropy every time.

That being said, I'm also okay with `generateByte` being the one without a default implementation. Most implementations work on byte-granularity anyways and for those that are extra careful about using entropy, they can implement `generateBool` too. For example, our default number generator probably doesn't need a special `generateBool`, but the secure one does.

packages/compiler/lib/src/compilation/hir/type.dart
```diff
@@ -125,6 +126,10 @@ abstract class CandyType with _$CandyType {
   static const implies =
       UserCandyType(ModuleId.coreOperatorsLogical, 'Implies');
 
+  // reflection
```
MarcelGarus:
```suggestion
  // random
```

MarcelGarus (changes requested):

packages/core/src/random/source.candy
JonasWanke: It's private already

packages/core/src/random/source.candy
JonasWanke: It's now changed to provide the three methods mentioned above, with `generateByte` being the default used by the other two.

MarcelGarus (approved):

PR #40, opened 2020-11-30
Implement map

MarcelGarus:


packages/core/src/collections/map/bucket_map.candy
```diff
@@ -7,66 +7,95 @@ use ...iterable
...
 class MapEntry<Key, Value> {
   let key: Key
   let hash: Int
-  let value: Value
+  mut let value: Value
 }
 
 class Bucket<Key: Equals & Hash, Value> {
-  let entries: List<MapEntry<Key, Value>>
+  /// An inefficient "mini-map" (it's typically pretty small).
```
JonasWanke: Maybe we should expose this class as a `LinkedHashMap`, or at least keep that in mind when we want to offer different map types in the future.

packages/core/src/collections/map/bucket_map.candy
```diff
@@ -7,66 +7,95 @@ use ...iterable
...
+  fun length(): Int { (entries_ as Iterable<Bucket<Key, Value>>).length() }
 
-  fun get(key: Key): Maybe<Value> {
-    let entries: Iterable<MapEntry<Key, Value>> = entries
-    entries.where({ it.key == key }).map<Value>({ it.value }).single()
+  fun get(key: Key): Maybe<Value> { entryForKey(key).map<Value>({ it.value }) }
+  fun entries(): Iterable<(Key, Value)> {
+    (entries_ as Iterable<MapEntry<Key, Value>>)
+      .map<(Key, Value)>({ Tuple(it.key, it.value) })
+    todo("Implement Bucket.entries")
```
JonasWanke: Doesn't this work already?

packages/core/src/collections/map/bucket_map.candy
```diff
@@ -7,66 +7,95 @@ use ...iterable
...
+    bucket.set(key, value)
   }
 
-  // TODO(marcelgarus): Implement `putIfAbsent` more efficiently than the default one. We can do
-  // this by only hashing the `key` once and caching the bucket.
-  // fun putIfAbsent(key: Key, ifAbsent: () => Value): Value
-
-  fun remove(key: Key): Bool {
-    todo("Implement BucketMap.remove")
+  fun putIfAbsent(key: Key, ifAbsent: () => Value): Value {
```
JonasWanke, line 38: This needs to return the value.

packages/core/src/collections/map/bucket_map.candy
```diff
@@ -7,66 +7,95 @@ use ...iterable
...
+    }
   }
 
-  fun clear() {
-    // TODO(marcelgarus): Implement
-    // for key in keys() {
-    //   remove(key)
-    // }
-    todo("Implement BucketMap.clear")
+  fun remove(key: Key): Bool {
```
JonasWanke: This needs to return whether an element was removed.

JonasWanke (changes requested):

JonasWanke (approved):

PR #44, opened 2020-12-16
Implement PetitParser

JonasWanke:


packages/petit_parser/src/parsers/custom.candy
```diff
@@ -0,0 +1,30 @@
+use ...parsers
+
+fun custom<Item, Output>(
+    length: Int,
+    converter: (List<Item>) => Maybe<Output>,
+    message: String
+  ): Parser<Item, Output> {
+  CustomParser<Item, Output>(length, converter, message)
+}
```
MarcelGarus:
```suggestion
fun custom<Item, Output>(
  length: Int,
  converter: (List<Item>) => Maybe<Output>,
  message: String
): Parser<Item, Output> {
  CustomParser<Item, Output>(length, converter, message)
}
```

MarcelGarus (changes requested):
An example per parser would help a lot at understanding what they're supposed to do.

MarcelGarus (approved):

PR #45, opened 2020-12-17
Data structures for the new AST

JonasWanke:
**Please merge #44 first!**

PR #46, opened 2020-12-17
Data structures for the new CST

JonasWanke:
**Please merge #45 first!**

### DIFF
--- a/packages/cst/src/types.candy
+++ b/packages/cst/src/types.candy
@@ -1,0 +1,84 @@
+use lexer
+use ..node
+
+public trait /* enum */ CstType
+
+
+public data class CstNamedType {
+  public let modifiers: List<CstNode<IdentifierToken>>
+  public let nameParts: List<CstNode<IdentifierToken>>
+  public let typeArguments: Maybe<CstNode<CstTypeArguments>>
+}
+impl CstNamedType: CstType
+
+
+public data class CstGroupType {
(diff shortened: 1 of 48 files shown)


packages/cst/src/types.candy
```diff
@@ -0,0 +1,84 @@
...
+  public let openingParenthesis: CstNode<PunctuationToken>
+  public let type: Maybe<CstNode<CstType>>
+  public let closingParenthesis: Maybe<CstNode<PunctuationToken>>
+}
+impl CstGroupType: CstType
+
+
+public data class CstFunctionType {
+  public let modifiers: List<CstNode<IdentifierToken>>
+  public let receiverType: Maybe<(CstNode<CstType>, Maybe<CstNode<PunctuationToken>>)>
```
MarcelGarus: There shouldn't be a dot. Function types with receivers looks like this:
```
keyword Foo (Int) -> Bool
```

MarcelGarus (changes requested):

packages/cst/src/types.candy
MarcelGarus:
```suggestion
  public let receiverType: Maybe<CstNode<CstType>>
```

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #49, opened 2020-12-30
Wrap native types & implement incremental framework

MarcelGarus:
Native types are now wrapped. So, there's `Bool`, `Int`, `String`, and `Array` now.
That also means that we don't need to use the native operators anymore: Calling the `add` function will also result in Dart code calling the `add` function – no magic `+` operator translation is necessary. Most importantly, all native types can now implement custom traits.

There's now the _incremental framework_, which can cache queries.

The implementation of the built-in types is not ugly, but repetitive. For example, there's no common way to get all the methods of all implemented traits, so that's something that's in every `builtin` class. We could refactor this into a reusable method, but I don't think it's worth the effort during bootstrapping.


packages/compiler_dart/lib/src/builtins.dart
```diff
@@ -122,9 +123,146 @@ class DartBuiltinCompiler extends BuiltinCompiler<dart.Spec> {
...
+                  ..returnType = t)),
+            ])
+            ..body = arrayT.call([
+              listT.property('generate').call([
+                dart.refer('length.value'),
+                // The Dart code generator doesn't support lambdas, so we do an ugly workaround.
+                dart.refer('(index) => generator').call([
+                  compileType(context, CandyType.int)
+                      .call([dart.refer('index')])
+                ]),
```
JonasWanke: It actually does: `dart.Method(…).closure` gives you a `dart.Expression` (which you already use at the end of this file). But we can leave it here and save the time.

packages/incremental/src/context.candy
```diff
@@ -0,0 +1,37 @@
...
+
+  static fun create(): QueryContext {
+    QueryContext(
+      MutableMap.empty<String, MutableMap<Equals & Hash, Equals & Hash>>(),
+      MutableList.empty<(String, Equals & Hash)>()
+    )
+  }
+
+  let cache: MutableMap<String, MutableMap<Equals & Hash, Equals & Hash>>
+  /// A cache from query names to a query-specific cache that maps from input to output.
```
JonasWanke, line 14: What about `MutableMap<(String, Equals & Hash), Equals & Hash>` – wouldn't that be simpler?

packages/incremental/src/query.candy
```diff
@@ -0,0 +1,32 @@
...
+  /// The input to the query.
+  ///
+  /// Multiple inputs can be represented as `Tuple`s or `List`s.
+
+  let runner: () => T
+  /// A runner that actually runs the query.
+  ///
+  /// The runner should be deterministic with regard to the inputs.
+
+  let shouldAlwaysRun: Bool
```
JonasWanke: Maybe we should rename this to a higher-level "isPure" or "isSideEffectFree". Also, this is currently ignored during execution.

JonasWanke (changes requested):
Incremental looks great, and the wrapping is very handy!

packages/incremental/src/context.candy
MarcelGarus: This won't work because `(Equals & Hash, Equals & Hash)` doesn't implement `Equals & Hash`.

packages/incremental/src/context.candy
```diff
@@ -17,6 +17,10 @@ class QueryContext {
   /// A stack of currently running queries. The tuples contain the name and input of the queries.
 
   fun run<T: Equals & Hash>(query: Query<T>): T {
+    if !query.isPure {
```
JonasWanke: We still have to check whether we produced a cycle

packages/compiler_dart/lib/src/builtins.dart
```diff
@@ -207,10 +207,13 @@ class DartBuiltinCompiler extends BuiltinCompiler<dart.Spec> {
...
-                dart.refer('(index) => generator').call([
-                  compileType(context, CandyType.int)
-                      .call([dart.refer('index')])
-                ]),
+                dart.Method((b) => b
+                  ..requiredParameters
+                      .add(dart.Parameter((b) => b..name = 'index'))
+                  ..body = dart.refer('generator').call([
+                    compileType(context, CandyType.int)
+                        .call([dart.refer('index')])
```
JonasWanke:
```suggestion
                    dart.refer('index')
```

JonasWanke (changes requested):

JonasWanke (approved):

PR #50, opened 2021-01-07
Add dart code builder

MarcelGarus:
This PR adds the dart_code_builder package and some missing functionality to core.

Of course, there's still Dart functionality missing (most importantly, utilities for common objects like the `@override` annotation etc.). Builders for files and expressions that handle allocating identifiers and dynamically adding imports would also be nice.
But since there's some changes in core and the bottom layer of the code building is already implemented, I think it's still a good time to merge.

### DIFF
--- a/packages/core/src/collections/iterable.candy
+++ b/packages/core/src/collections/iterable.candy
@@ -248,8 +248,12 @@ public trait Iterable<Item> {
 
   fun join(separator: String): String {
     mut let s = ""
+    mut let isFirst = true
+
     for item in this {
-      if !(s == "") {
+      if isFirst {
+        isFirst = false
+      } else {
         s = "{s}{separator}"
       }
       s = "{s}{item}"
--- a/packages/core/src/collections/iterable.candy
+++ b/packages/core/src/collections/iterable.candy
@@ -2,6 +2,7 @@ use ...bool
 use ...operators
 use ...maybe
 use ...numbers
+use ...string
 use ...todo
 use ..list
 
@@ -209,6 +210,10 @@ public trait Iterable<Item> {
   // TODO(marcelgarus): reversed
   // TODO(marcelgarus): zip
 
+  fun followedBy(other: Iterable<Item>): Iterable<Item> {
+    FollowedByIterable<Item>(this, other)
+  }
+
   // Conversions.
 
   fun toList(): List<Item> {
@@ -240,6 +245,17 @@ public trait Iterable<Item> {
   // TODO(marcelgarus): for Iterable<Iterable<T>>: flatten
   // TODO(marcelgarus): for Iterable<Number>: average, sum, min, max, minBy, maxBy
   // TODO(marcelgarus): for Iterable<Maybe<T>>: whereSome, unwrapAll
+
+  fun join(separator: String): String {
+    mut let s = ""
+    for item in this {
+      if !(s == "") {
+        s = "{s}{separator}"
+      }
+      s = "{s}{item}"
+    }
+    s
+  }
 }
 
 class MappedIterable<In, Out> {
@@ -288,6 +304,29 @@ impl<Item> WhereIterator<Item>: Iterator<Item> {
         }
       }
     }
+  }
+}
+
+class FollowedByIterable<Item> {
+  let parent: Iterable<Item>
+  let following: Iterable<Item>
+}
+impl<Item> FollowedByIterable<Item>: Iterable<Item> {
+  fun iterator(): Iterator<Item> {
+    FollowedByIterator<Item>(parent.iterator(), following.iterator())
+  }
+}
+class FollowedByIterator<Item> {
+  let parent: Iterator<Item>
+  let following: Iterator<Item>
+}
+impl<Item> FollowedByIterator<Item>: Iterator<Item> {
+  fun next(): Maybe<Item> {
+    let a = parent.next()
+    if a is Some {
+      return a
+    }
+    following.next()
   }
 }
 
--- a/packages/core/src/collections/list/module.candy
+++ b/packages/core/src/collections/list/module.candy
@@ -47,24 +47,6 @@ public trait List<Item>: Iterable<Item> {
     }
     result
   }
-
-  fun followedBy(other: List<Item>): List<Item> {
-    let resultLength = length() + (other as Iterable<Item>).length()
-    let result = ArrayList.create<Item>(resultLength)
-
-    mut let index = 0
-    while index < length() {
-      (result as MutableList<Item>).append(get(index).unwrap() as Item)
-      index = index + 1
-    }
-
-    mut let index = 0
-    while index < (other as Iterable<Item>).length() {
-      (result as MutableList<Item>).append((other as Iterable<Item>).get(index).unwrap() as Item)
-      index = index + 1
-    }
-    result
-  }
 }
 
 public trait MutableList<Item>: List<Item> {
--- a/packages/core/src/collections/list/module.candy
+++ b/packages/core/src/collections/list/module.candy
@@ -48,7 +48,7 @@ public trait List<Item>: Iterable<Item> {
     result
   }
 
-  fun followedBy(other: List<Item>): List<Item> {
+  fun followedByList(other: List<Item>): List<Item> {
     let resultLength = length() + (other as Iterable<Item>).length()
     let result = ArrayList.create<Item>(resultLength)
 
--- a/packages/core/src/collections/list/module.candy
+++ b/packages/core/src/collections/list/module.candy
@@ -47,6 +47,24 @@ public trait List<Item>: Iterable<Item> {
     }
     result
   }
+
+  fun followedBy(other: List<Item>): List<Item> {
+    let resultLength = length() + (other as Iterable<Item>).length()
+    let result = ArrayList.create<Item>(resultLength)
+
+    mut let index = 0
+    while index < length() {
+      (result as MutableList<Item>).append(get(index).unwrap() as Item)
+      index = index + 1
+    }
+
+    mut let index = 0
+    while index < (other as Iterable<Item>).length() {
+      (result as MutableList<Item>).append((other as Iterable<Item>).get(index).unwrap() as Item)
+      index = index + 1
+    }
+    result
+  }
 }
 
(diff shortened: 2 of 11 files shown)


packages/core/src/collections/iterable.candy
```diff
@@ -240,6 +245,17 @@ public trait Iterable<Item> {
   // TODO(marcelgarus): for Iterable<Iterable<T>>: flatten
   // TODO(marcelgarus): for Iterable<Number>: average, sum, min, max, minBy, maxBy
   // TODO(marcelgarus): for Iterable<Maybe<T>>: whereSome, unwrapAll
+
+  fun join(separator: String): String {
+    mut let s = ""
+    for item in this {
+      if !(s == "") {
+        s = "{s}{separator}"
+      }
```
JonasWanke: This skips one separator if the first element's `toString()` returns an empty string.

packages/core/src/collections/list/module.candy
```diff
@@ -47,24 +47,6 @@ public trait List<Item>: Iterable<Item> {
...
-      index = index + 1
-    }
-
-    mut let index = 0
-    while index < (other as Iterable<Item>).length() {
-      (result as MutableList<Item>).append((other as Iterable<Item>).get(index).unwrap() as Item)
-      index = index + 1
-    }
-    result
-  }
```
JonasWanke: Please re-add this function

packages/dart_code_builder/src/example.candy
```diff
@@ -0,0 +1,140 @@
...
+        namedParameters = List.of1<Parameter>(Parameter(
+          "blub",
+          Some<DartType>(UserType("int", List.empty<DartType>(), prefix = None<String>())),
+          defaultValue = None<Expression>(),
+        )),
+        body = None<Body>(),
+      ),
+    ),
+  ))
+  let file = File(list)
```
JonasWanke: The Dart analyzer calls this a `CompilationUnit`, which has `directives` and `declarations` as separate lists because directives may not appear after a declaration.

packages/dart_code_builder/src/example.candy
```diff
@@ -0,0 +1,140 @@
+use ..nodes
+
+fun main() {
+  let list = MutableList.empty<Directive | Declaration>()
+  list.append(Import(
+    "package:foo/foo.dart",
+    alias = Some<String>("foo"),
```
JonasWanke: The Dart analyzer calls this a `prefix`

packages/dart_code_builder/src/example.candy
```diff
@@ -0,0 +1,140 @@
...
+        (UserType(
+          "Cached",
+          parameters = List.empty<DartType>(),
+          prefix = None<String>(),
+        ) as Expression).call(
+          positionalArguments = List.empty<Expression>(),
+          namedArguments = Map.empty<String, Expression>(),
+          typeArguments = List.empty<DartType>(),
+        )
+      )
```
JonasWanke: Annotations in Dart can only be of the form `'@' Identifier ('.' SimpleIdentifier)? ArgumentList?`. More specifically, you cannot simply use any expression and not even a function call with type arguments.

https://pub.dev/documentation/analyzer/latest/dart_ast_ast/Annotation-class.html

packages/dart_code_builder/src/example.candy
```diff
@@ -0,0 +1,140 @@
...
+    ),
+    _returns = Some<DartType>(UserType(
```
JonasWanke: The Dart analyzer calls this a `NamedType`

packages/dart_code_builder/src/example.candy
```diff
@@ -0,0 +1,140 @@
...
+      "Foo",
+      parameters = List.empty<DartType>(),
```
JonasWanke: These are `typeArguments`

packages/dart_code_builder/src/example.candy
```diff
@@ -0,0 +1,140 @@
...
+      prefix = Some<String>("foo"),
```
JonasWanke: Dart combines name and prefix into an [`Identifier`](https://pub.dev/documentation/analyzer/latest/dart_ast_ast/Identifier-class.html)

packages/dart_code_builder/src/example.candy
```diff
@@ -0,0 +1,140 @@
...
+    implements_ = List.of3<DartType>(
+      UserType("Baz", List.empty<DartType>(), None<String>()),
+      UserType("Wheez", List.empty<DartType>(), None<String>()),
+      UserType("Fnop", List.empty<DartType>(), None<String>()),
+    ),
+    with_ = List.of2<DartType>(
+      UserType("Whup", List.empty<DartType>(), None<String>()),
+      UserType("Flang", List.empty<DartType>(), None<String>()),
+    ),
+    body = List.of4<Field | Getter | Setter | Function>(
```
JonasWanke: Members also include `Constructor`s

packages/dart_code_builder/src/example.candy
```diff
@@ -0,0 +1,140 @@
...
+      // static const numBlubs = 123;
+      Field("numBlubs", isStatic = true, mutability = Const(), Some<Expression>(Literal(123))),
```
JonasWanke: To my knowledge, only the following modifier combinations are valid:
* \<nothing\>
* `final`
* `late`
* `late final`
* `external`
* `external static`
* `external covariant`
* `covariant`
* `covariant late`
* `static`
* `static final`
* `static const`
* `static late`
* `static late final`

https://pub.dev/documentation/analyzer/latest/dart_ast_ast/FieldDeclaration-class.html

packages/dart_code_builder/src/example.candy
```diff
@@ -0,0 +1,140 @@
...
+        "isEmpty",
+        parameter = Parameter(
+          "value",
+          type = Some<DartType>(UserType("bool", List.empty<DartType>(), None<String>())),
+          defaultValue = None<Expression>(),
+        ),
+        body = Some<Body>(BlockBody(List.of1<Statement>(
+          If(
+            Identifier("value", None<String>()),
+            then = List.of1<Statement>(
```
JonasWanke: In the Dart analyzer, this is a single statement which can be a `Block`.

* https://pub.dev/documentation/analyzer/latest/dart_ast_ast/IfStatement-class.html
* https://pub.dev/documentation/analyzer/latest/dart_ast_ast/Statement-class.html
* https://pub.dev/documentation/analyzer/latest/dart_ast_ast/Block-class.html

JonasWanke (changes requested):
Not covered:

* null literal
* string literals with interpolation
* double literals
* mixins
* extensions
* cascades
* constructors
* some expressions:
  * await
  * binary: assignments, `as`, `is`, `is!`, `!==`, `^`, `===`, `>>`, `>>>`, `<<`, `?.`
  * prefix operator: `~`
  * ternary
  * lambdas
* function declarations in bodies
* continue
* for, do-while
* switch
* try/catch, throw

MarcelGarus:
Just to clarify: This package is not supposed to be a one to one port of the Dart AST – that would be quite a lot of work with only limited benefits. For now, its only purpose is to be a wrapper allowing us to generate Dart code needed for the compiler.

_Maybe_ in the future it _may_ make sense to more strictly follow the Dart guidelines, but this is a pragmatic, very rough approximation of Dart syntax. More specifically, there currently are valid Dart expressions that can't be produced (like extension methods) and invalid Dart expressions that can be produced (like `Identifier("a * (3 + ")`).

However, some of the things you mentioned actually are needed for our compiler, so I added them:
- null literal
- mixins
- constructors
  - initializing formals
- as
- is
- lambdas
- continue

There are still some things missing (`throw`, null safety features), but I think this is enough to get us started. We can add features as we go.

packages/dart_code_builder/src/to_code.candy
```diff
@@ -0,0 +1,411 @@
...
+
+impl Part: ToCode {
+  fun toCode(): String { "part '{path}';" }
+}
+
+impl PartOf: ToCode {
+  fun toCode(): String { "part of '{path}';" }
+}
+
+let a = 3
```
JonasWanke: Does this have any purpose?

packages/dart_code_builder/src/to_code.candy
```diff
@@ -0,0 +1,411 @@
...
+        .join(", ")
+      s = "{s} implements {implementsCode}"
+    }
+    if (with_ as Iterable).isNotEmpty() {
+      let withCode = (with_ as Iterable<DartType>)
+        .map<String>({ (it as ToCode).toCode() })
+        .join(", ")
+      s = "{s} with {withCode}"
+    }
+    let bodyCode = (body as Iterable<Field | Getter | Setter | Function>)
```
JonasWanke:
```suggestion
    let bodyCode = (body as Iterable<Constructor | Field | Getter | Setter | Function>)
```

packages/dart_code_builder/src/to_code.candy
```diff
@@ -0,0 +1,411 @@
...
+    "{s}(parameters)"
+  }
+}
+
+impl Expression: ToCode
+impl Identifier: ToCode {
+  fun toCode(): String {
+    mut let s = name
+    if prefix is Some {
+      print("f")
```
JonasWanke: Can this be removed?

packages/dart_code_builder/src/to_code.candy
```diff
@@ -0,0 +1,411 @@
+use ..nodes
+
+fun indent(code: String, indentFirstLine: Bool): String {
```
JonasWanke: `indentFirstLine` is always `true`.

JonasWanke (changes requested):

packages/dart_code_builder/src/to_code.candy
MarcelGarus: Nope, just for debugging and seeing if the HIR gets available for type hints 😅

packages/dart_code_builder/src/to_code.candy
MarcelGarus: Oh, I was wondering why it prints an `f` 😁

JonasWanke (approved):

PR #51, opened 2021-01-10
Parser

JonasWanke:


packages/core/src/collections/list/module.candy
```diff
@@ -24,6 +24,27 @@ public trait List<Item>: Iterable<Item> {
...
+  }
+  static fun of10<Item>(first: Item, second: Item, third: Item, fourth: Item, fifth: Item, sixth: Item, seventh: Item, eigth: Item, ninth: Item, tenth: Item): List<Item> {
+    ArrayList.of10<Item>(first, second, third, fourth, fifth, sixth, seventh, eigth, ninth, tenth)
+  }
+  static fun of11<Item>(first: Item, second: Item, third: Item, fourth: Item, fifth: Item, sixth: Item, seventh: Item, eigth: Item, ninth: Item, tenth: Item, eleventh: Item): List<Item> {
+    ArrayList.of11<Item>(first, second, third, fourth, fifth, sixth, seventh, eigth, ninth, tenth, eleventh)
+  }
+  static fun of12<Item>(first: Item, second: Item, third: Item, fourth: Item, fifth: Item, sixth: Item, seventh: Item, eigth: Item, ninth: Item, tenth: Item, eleventh: Item, twelfth: Item): List<Item> {
+    ArrayList.of12<Item>(first, second, third, fourth, fifth, sixth, seventh, eigth, ninth, tenth, eleventh, twelfth)
+  }
```
MarcelGarus, line 47: For what exactly do we need this abomination?

packages/cst/src/expressions.candy
```diff
@@ -1,23 +1,18 @@
...
-
-public data class CstIntExpression {
-  public let value: CstNode<LiteralToken<Int>>
+public class CstIntExpression {
+  public let value: CstNode<IntLiteralToken>
 }
 impl CstIntExpression: CstExpression
 
 
-public data class CstStringExpression {
```
MarcelGarus, line 20: Why are all data keywords removed? Because `toString` is now implemented by default?

MarcelGarus, line 20: Why are all `data` keywords removed? Because `toString` is now implemented by default?

packages/cst/src/parser.candy
```diff
@@ -72,49 +83,881 @@ class CstParser {
...
+  let type = notYetDefined<String, CstNode<CstType>>("Parser not yet defined.")
+  fun initType() {
+    let closingParenthesisOrOtherEndDetection = punctuationOrEnd(")")
+        .or<Maybe<CstNode<PunctuationToken>>>(
+          punctuation(">")
+              .or<CstNode<PunctuationToken>>(punctuation("}"))
+              .and()
+              .map<Maybe<CstNode<PunctuationToken>>>({ None<CstNode<PunctuationToken>>() })
+        )
+        .map<Maybe<CstNode<PunctuationToken>>>({ it as Maybe<CstNode<PunctuationToken>> })
```
MarcelGarus: Would a `cast` method make sense? I saw this pattern a couple of times already.

packages/petit_parser/src/parsers/miscellaneous/any.candy
```diff
@@ -0,0 +1,20 @@
...
+public fun any<Item>(message: String = "Expected any input."): Parser<Item, Item> {
+  AnyParser<Item>(message)
+}
+
+class AnyParser<Item> {
+  let message: String
+}
+impl<Item> AnyParser<Item>: Parser<Item, Item> {
+  fun parse(context: Context<Item>): Result<Success<Item>, Failure> {
+    let buffer = (context.buffer as Iterable<Item>)
```
MarcelGarus, line 12:
```suggestion
    let buffer = context.buffer as Iterable<Item>
```

packages/petit_parser/src/parsers/miscellaneous/any.candy
```diff
@@ -0,0 +1,20 @@
...
+    if (context.offset >= buffer.length()) {
```
MarcelGarus:
```suggestion
    if context.offset >= buffer.length() {
```

packages/petit_parser/src/parsers/module.candy
```diff
@@ -78,20 +101,34 @@ public trait Parser<Item, Output> {
...
+              .sequence<Output>(this)
+              .repeat(minimumItems - 1, None<Int>())
+              .map<List<(SeparatorOutput, Output)>>({ it as List<(SeparatorOutput, Output)> })
+        )
+        // .map<(Output, List<Output | SeparatorOutput>)>({
+        //   let a = it
+        //   let initial = a.first
+        //   let sequence = a.second
+        //   it as (Output, List<Output | SeparatorOutput>)
+        // })
```
MarcelGarus: We should remove the comments or add a todo comment

MarcelGarus (changes requested):

packages/core/src/collections/list/module.candy
JonasWanke: The parser uses `of12` for a list of supported keywords. The others up to `of10` were formerly used by the parser as well, but due to a big structural change, are no longer required there. But as we already have them and they will be removed when we have collection literals, I'd recommend we just leave them there in the meantime.

packages/petit_parser/src/parsers/miscellaneous/any.candy
JonasWanke: Unfortunately, our old parser has problems without the parentheses…

packages/cst/src/expressions.candy
JonasWanke: `CstNode<T: Equals & Hash>` can't currently become a data class because its `trailingTrivia: List<WhitespaceToken | CommentToken>` currently doesn't implement `Hash & Equals`

MarcelGarus (approved):

PR #56, opened 2021-01-12
CST-AST-lowering

JonasWanke:


packages/ast/src/types.candy
```diff
@@ -1,63 +1,83 @@
...
 use ..node
 
-public trait /* enum */ CandyType {
+public class AstTypeId {
+  let parentId: AstDeclarationId | AstExpressionId
+  let value: Int
+}
+impl AstTypeId: AstNodeId
+
+public trait /* enum */ AstCandyType {
```
MarcelGarus: Well, with the prefix, the name doesn't clash with the built-in type anymore. So, we can just call it `AstType`.

packages/ast/src/errors/missing_expression_in_interpolation.candy
```diff
@@ -0,0 +1,17 @@
+use compiler_utils
+use cst
+
+public class MissingExpressionInInterpolationCompilerError {
+  public let file: File
+  public let stringPart: CstNode<CstInterpolatedStringPart>
+}
+impl MissingExpressionInInterpolationCompilerError: CompilerError {
+  public fun id(): String { "missingExpressionInInterpolation" }
```
MarcelGarus: Should we use lower kebab for error ids? Makes them more distinct and IMO skimmable: `missing-expression-in-interpolation` vs `missingExpressionInInterpolation`.

packages/ast/src/lowering.candy
```diff
@@ -0,0 +1,339 @@
...
+  let id = AstUseLineId(fileId, useLineIndex)
+
+  let modifiers = lowerModifiers(context, id, cst.child.modifiers)
+  let target = cst.child.target
+      .map<AstUseLineTarget>({
+        if it.child is CstLocalRelativeUseLineTarget {
+          lowerLocalRelativeUseLineTarget(context, id, it.cast<CstLocalRelativeUseLineTarget>())
+        } else {
+          lowerGlobalUseLineTarget(context, id, it.cast<CstGlobalUseLineTarget>())
+        }
```
MarcelGarus, line 68: Fine for now. In the future, instead of many top-level functions, we should probably have a `LowerToAst` trait or something similar that is implemented for `CstNode<CstFile>`, `Cst<Node<CstUseLine>>` etc. so we can just to `it.toAst()`

packages/ast/src/lowering.candy
```diff
@@ -0,0 +1,339 @@
...
+      } else {
+        punctuationExpectedNext = false
+      }
+    } else {
+      if punctuationExpectedNext {
+        /// TODO(JonasWanke)
+      } else {
+        let identifierId = AstIdentifierId(id, (packagePath as Iterable<AstIdentifier>).length())
+        packagePath.append(
+          AstIdentifier(identifierId, (node.cast<IdentifierToken>()).child.identifier),
```
MarcelGarus:
```suggestion
          AstIdentifier(identifierId, node.cast<IdentifierToken>().child.identifier),
```

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #57, opened 2021-01-20
Type solving

MarcelGarus:
This PR is the first step towards a good type solver: type unification. Type unification refers to testing if two types can be the same.

This doesn't work directly on Candy types, but rather on a simplified version – `SolverType`s – that Candy types are later lowered to. For example, a tuple type might be lowered into a `Tuple` type with type parameters that is treated like any other generic type.

Contrary to Rust, we also support union and intersection types, so we'll have to think about how to deal with that.

Here are some examples of things that already work (output of the example program, type parameters are prefixed with a question mark):

```
Unifying String and String: possible, no substitutions necessary
Unifying Int and String: not possible
Unifying Int and ?A: possible, with these substitutions: ?A = Int
Unifying List<?A> and List<Set<?B>>: possible, with these substitutions: ?A = Set<?B>
Unifying Map<?A, ?A> and Map<Int, String>: not possible
Unifying Map<?A, ?A> and Map<Int, Int>: possible, with these substitutions: ?A = Int
Unifying Map<?A, ?A> and Map<?B, Int>: possible, with these substitutions: ?A = Int, ?B = Int
Unifying Set<?A> and Set<?B>: possible, with these substitutions: ?A = ?B
Unifying ?A and List<?A>: not possible
Unifying Map<?A, ?B> and Map<List<?B>, List<?A>>: not possible
```


packages/hir/src/expressions.candy
```diff
@@ -1,182 +1,178 @@
...
+impl HirStringExpression: HirExpression
 
-trait StringPart
-impl StringPart: Expression
+trait HirStringPart
 
-public class LiteralStringPart {
-  public let uri: ExpressionUri<LiteralStringPart>
+public class HirLiteralStringPart {
+  public let uri: HirExpressionUri<HirLiteralStringPart>
```
JonasWanke: This won't work if `HirStringPart` doesn't implement `HirExpression`. Maybe we should inline string parts instead of using URIs for these?

packages/hir/src/solver/module.candy
```diff
@@ -0,0 +1 @@
+use types
```
JonasWanke:
```suggestion
public use types
```

packages/hir/src/solver/types.candy
```diff
@@ -0,0 +1,220 @@
+
```
JonasWanke:
```suggestion
```

packages/hir/src/solver/types.candy
```diff
@@ -0,0 +1,220 @@
...
+      }
+    } else {
+      SolverValue(
+        (this as SolverValue).name,
+        ((this as SolverValue).parameters as Iterable<SolverType>)
+          .map<SolverType>({ it.substitute(variable, substitution) })
+          .toList()
+      )
+    }
+  }
```
JonasWanke: Could you move these implementations to the respective subclasses?

packages/hir/src/solver/types.candy
```diff
@@ -0,0 +1,220 @@
...
+          return None<Map<SolverVariable, SolverType>>()
+        }
+      }
+    }
+
+    Some<Map<SolverVariable, SolverType>>(solutionSpace)
+  }
+}
+impl SolverType: Hash {
+  fun hash<T>(hasher: Hasher<T>) { assert(false, "subclasses should override this") }
```
JonasWanke: Can't we just remove this method then and declare `trait SolverType: Hash & Equals`

packages/hir/src/solver/types.candy
```diff
@@ -0,0 +1,220 @@
...
+}
+
+class SolverValue {
+  /// A concrete type. May contain `SolverVariable`s inside its `parameters`.
+  ///
+  /// Examples:
+  /// * `List<?T>`
+  /// * `String`
+  /// * `Set<Int>`
+  /// * `Tuple<Int, String>`
```
JonasWanke:
```suggestion
  /// Examples:
  ///
  /// * `List<?T>`
  /// * `String`
  /// * `Set<Int>`
  /// * `Tuple<Int, String>`
```

JonasWanke (changes requested):

JonasWanke (approved):

PR #58, opened 2021-01-21
CST-AST-lowering of types and declarations

JonasWanke:

### DIFF
--- a/packages/ast/src/declarations.candy
+++ b/packages/ast/src/declarations.candy
@@ -20,7 +20,7 @@ public class AstModule {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
   public let modifiers: List<AstIdentifier>
-  public let name: Maybe<AstIdentifier>
+  public let name: AstIdentifier
   public let innerDeclarations: List<
     AstModule
     | AstTrait
@@ -37,7 +37,7 @@ public class AstTrait {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
   public let modifiers: List<AstIdentifier>
-  public let name: Maybe<AstIdentifier>
+  public let name: AstIdentifier
   public let typeParameters: List<AstTypeParameter>
   public let upperBound: Maybe<AstType>
   public let innerDeclarations: List<AstTrait | AstClass | AstFunction | AstProperty>
@@ -59,7 +59,7 @@ public class AstClass {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
   public let modifiers: List<AstIdentifier>
-  public let name: Maybe<AstIdentifier>
+  public let name: AstIdentifier
   public let typeParameters: List<AstTypeParameter>
   public let innerDeclarations: List<AstTrait | AstClass | AstFunction | AstProperty>
 }
@@ -70,7 +70,7 @@ public class AstFunction {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
   public let modifiers: List<AstIdentifier>
-  public let name: Maybe<AstIdentifier>
+  public let name: AstIdentifier
   public let typeParameters: List<AstTypeParameter>
   public let valueParameters: List<AstValueParameter>
   public let returnType: Maybe<AstType>
@@ -88,7 +88,7 @@ impl AstValueParameterId: AstNodeId
 public class AstValueParameter {
   public let id: AstValueParameterId
   public let modifiers: List<AstIdentifier>
-  public let name: Maybe<AstIdentifier>
+  public let name: AstIdentifier
   public let type: Maybe<AstType>
   public let defaultValue: Maybe<AstExpression>
 }
@@ -99,7 +99,7 @@ public class AstProperty {
   public let candyDoc: Maybe<AstCandyDoc>
   public let modifiers: List<AstIdentifier>
   public let isMutable: Bool
-  public let name: Maybe<AstIdentifier>
+  public let name: AstIdentifier
   public let type: Maybe<AstType>
   public let initializer: Maybe<AstExpression>
 }
--- a/packages/ast/src/declarations.candy
+++ b/packages/ast/src/declarations.candy
@@ -59,7 +59,7 @@ public class AstClass {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
   public let modifiers: List<AstIdentifier>
-  public let name: AstIdentifier
+  public let name: Maybe<AstIdentifier>
   public let typeParameters: List<AstTypeParameter>
   public let innerDeclarations: List<AstTrait | AstClass | AstFunction | AstProperty>
 }
--- a/packages/ast/src/declarations.candy
+++ b/packages/ast/src/declarations.candy
@@ -70,7 +70,7 @@ public class AstFunction {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
   public let modifiers: List<AstIdentifier>
-  public let name: AstIdentifier
+  public let name: Maybe<AstIdentifier>
   public let typeParameters: List<AstTypeParameter>
   public let valueParameters: List<AstValueParameter>
   public let returnType: Maybe<AstType>
@@ -108,7 +108,6 @@ impl AstProperty: AstDeclaration
 
 public class AstBodyId {
   let parentId: AstDeclarationId
-  let name: String
 }
 impl AstBodyId: AstNodeId
 
--- a/packages/ast/src/declarations.candy
+++ b/packages/ast/src/declarations.candy
@@ -48,7 +48,7 @@ public class AstImpl {
   public let id: AstDeclarationId
   public let modifiers: List<AstIdentifier>
   public let typeParameters: List<AstTypeParameter>
-  public let type: Maybe<AstType>
+  public let type: AstType
   public let traits: List<AstNamedType>
   public let innerDeclarations: List<AstFunction | AstProperty>
 }
--- a/packages/ast/src/declarations.candy
+++ b/packages/ast/src/declarations.candy
@@ -48,8 +48,8 @@ public class AstImpl {
   public let id: AstDeclarationId
   public let modifiers: List<AstIdentifier>
   public let typeParameters: List<AstTypeParameter>
-  public let type: AstType
-  public let traits: List<AstType>
+  public let type: Maybe<AstType>
+  public let traits: List<AstNamedType>
   public let innerDeclarations: List<AstFunction | AstProperty>
 }
 impl AstImpl: AstDeclaration
--- a/packages/ast/src/declarations.candy
+++ b/packages/ast/src/declarations.candy
@@ -11,7 +11,7 @@ public class AstDeclarationId {
 impl AstDeclarationId: AstNodeId
 
 public trait /* enum */ AstDeclaration {
-  public let modifiers: List<String>
+  public let modifiers: List<AstIdentifier>
 }
 impl AstDeclaration: AstNode
 
@@ -19,8 +19,8 @@ impl AstDeclaration: AstNode
 public class AstModule {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
-  public let modifiers: List<String>
-  public let name: Maybe<String>
+  public let modifiers: List<AstIdentifier>
+  public let name: Maybe<AstIdentifier>
   public let innerDeclarations: List<
     AstModule
     | AstTrait
@@ -36,8 +36,8 @@ impl AstModule: AstDeclaration
 public class AstTrait {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
-  public let modifiers: List<String>
-  public let name: Maybe<String>
+  public let modifiers: List<AstIdentifier>
+  public let name: Maybe<AstIdentifier>
   public let typeParameters: List<AstTypeParameter>
   public let upperBound: Maybe<AstType>
   public let innerDeclarations: List<AstTrait | AstClass | AstFunction | AstProperty>
@@ -46,7 +46,7 @@ impl AstTrait: AstDeclaration
 
 public class AstImpl {
   public let id: AstDeclarationId
-  public let modifiers: List<String>
+  public let modifiers: List<AstIdentifier>
   public let typeParameters: List<AstTypeParameter>
   public let type: AstType
   public let traits: List<AstType>
@@ -58,8 +58,8 @@ impl AstImpl: AstDeclaration
 public class AstClass {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
-  public let modifiers: List<String>
-  public let name: String
+  public let modifiers: List<AstIdentifier>
+  public let name: AstIdentifier
   public let typeParameters: List<AstTypeParameter>
   public let innerDeclarations: List<AstTrait | AstClass | AstFunction | AstProperty>
 }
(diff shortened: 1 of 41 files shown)


packages/ast/src/declarations.candy
```diff
@@ -46,10 +46,10 @@ impl AstTrait: AstDeclaration
 
 public class AstImpl {
   public let id: AstDeclarationId
-  public let modifiers: List<String>
+  public let modifiers: List<AstIdentifier>
   public let typeParameters: List<AstTypeParameter>
-  public let type: AstType
-  public let traits: List<AstType>
+  public let type: Maybe<AstType>
```
MarcelGarus: Don't you always have to provide a type?

packages/ast/src/declarations.candy
```diff
@@ -11,16 +11,16 @@ public class AstDeclarationId {
...
 impl AstDeclaration: AstNode
 
 
 public class AstModule {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
-  public let modifiers: List<String>
-  public let name: Maybe<String>
+  public let modifiers: List<AstIdentifier>
+  public let name: Maybe<AstIdentifier>
```
MarcelGarus: Doesn't a module always need a name?

packages/ast/src/declarations.candy
```diff
@@ -36,8 +36,8 @@ impl AstModule: AstDeclaration
 public class AstTrait {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
-  public let modifiers: List<String>
-  public let name: Maybe<String>
+  public let modifiers: List<AstIdentifier>
+  public let name: Maybe<AstIdentifier>
```
MarcelGarus: A trait should always have a name

packages/ast/src/declarations.candy
```diff
@@ -46,10 +46,10 @@ impl AstTrait: AstDeclaration
 
 public class AstImpl {
   public let id: AstDeclarationId
-  public let modifiers: List<String>
+  public let modifiers: List<AstIdentifier>
   public let typeParameters: List<AstTypeParameter>
-  public let type: AstType
-  public let traits: List<AstType>
+  public let type: Maybe<AstType>
+  public let traits: List<AstNamedType>
```
MarcelGarus, line 52: Didn't we decide on disallowing multiple (comma-separated) types in favor of intersection types?
```
impl Foo: Bar & Baz
```
instead of 
```
impl Foo: Bar, Baz
```

packages/ast/src/declarations.candy
```diff
@@ -58,8 +58,8 @@ impl AstImpl: AstDeclaration
 public class AstClass {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
-  public let modifiers: List<String>
-  public let name: String
+  public let modifiers: List<AstIdentifier>
+  public let name: Maybe<AstIdentifier>
```
MarcelGarus: Should always have a name

packages/ast/src/errors/duplicate_declaration_name.candy
```diff
@@ -0,0 +1,26 @@
...
+  // TODO(JonasWanke): highlight previous occurrences of that name
+}
+impl DuplicateDeclarationNameCompilerError: CompilerError {
+  public fun id(): String { "duplicate-declaration-name" }
+
+  fun span(): Span {
+    if node.child is CstModule { return node.cast<CstModule>().child.name.unwrap().child.span }
+    if node.child is CstTrait { return node.cast<CstTrait>().child.name.unwrap().child.span }
+    if node.child is CstClass { return node.cast<CstClass>().child.name.unwrap().child.span }
+    if node.child is CstProperty { return node.cast<CstProperty>().child.name.unwrap().child.span }
```
MarcelGarus: Okay for now. In the long term, a `HasName` trait or something like that would be useful.

packages/ast/src/declarations.candy
```diff
@@ -69,8 +69,8 @@ impl AstClass: AstDeclaration
 public class AstFunction {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
-  public let modifiers: List<String>
-  public let name: String
+  public let modifiers: List<AstIdentifier>
+  public let name: Maybe<AstIdentifier>
```
MarcelGarus: Functions always have a name

packages/ast/src/declarations.candy
```diff
@@ -87,27 +87,27 @@ impl AstValueParameterId: AstNodeId
 
 public class AstValueParameter {
   public let id: AstValueParameterId
-  public let name: Maybe<String>
+  public let modifiers: List<AstIdentifier>
+  public let name: Maybe<AstIdentifier>
```
MarcelGarus: Value parameters always have a name

packages/ast/src/declarations.candy
```diff
@@ -87,27 +87,27 @@ impl AstValueParameterId: AstNodeId
...
 
 
 public class AstProperty {
   public let id: AstDeclarationId
   public let candyDoc: Maybe<AstCandyDoc>
-  public let modifiers: List<String>
+  public let modifiers: List<AstIdentifier>
   public let isMutable: Bool
-  public let name: Maybe<String>
+  public let name: Maybe<AstIdentifier>
```
MarcelGarus: Properties always have a name

packages/ast/src/errors/invalid_declaration_type.candy
```diff
@@ -0,0 +1,39 @@
...
+    }
+    if node.child is CstClass {
+      return node.cast<CstClass>().child.classKeyword.child.span
+    }
+    if node.child is CstFunction {
+      return node.cast<CstFunction>().child.functionKeyword.child.span
+    }
+    if node.child is CstProperty {
+      return node.cast<CstProperty>().child.keyword.child.span
+    }
```
MarcelGarus: A `HasKeyword` trait would be useful. We could simply do `node.cast<HasKeyword>().child.keyword().child.span`.

packages/ast/src/errors/missing_function_value_parameters.candy
```diff
@@ -0,0 +1,16 @@
...
+
+public class MissingValueParametersCompilerError {
+  public let file: File
+  public let node: CstNode<CstFunction>
+}
+impl MissingValueParametersCompilerError: CompilerError {
+  public fun id(): String { "missing-function-value-parameters-interpolation" }
+
+  public fun location(): Location { Location(file, node.child.functionKeyword.child.span) }
+  public fun title(): String { "Missing function value parameters string interpolation." }
```
MarcelGarus:
```suggestion
  public fun title(): String { "Missing function value parameters." }
```

packages/ast/src/types.candy
```diff
@@ -35,49 +41,48 @@ impl AstFunctionType: AstType
...
   let name: String
 }
 impl AstTypeParameterId: AstNodeId
 
 public class AstTypeParameter {
   public let id: AstTypeParameterId
-  public let modifiers: List<String>
-  public let name: Maybe<String>
+  public let modifiers: List<AstIdentifier>
+  public let name: Maybe<AstIdentifier>
```
MarcelGarus, line 74: Always has a name

packages/ast/src/types.candy
```diff
@@ -35,49 +41,48 @@ impl AstFunctionType: AstType
...
 }
 
 
 public class AstTypeArgumentId {
-  let declaration: AstDeclarationId
-  let name: String
+  let parentId: AstTypeId | AstExpressionId
+  let index: Int
 }
 impl AstTypeArgumentId: AstNodeId
```
MarcelGarus, line 83: What exactly do we need this for? Aren't the `AstType`s inlined in the declaration?

packages/core/src/collections/map/module.candy
```diff
@@ -49,6 +49,11 @@ public trait MutableMap<Key: Equals & Hash, Value>: Map<Key, Value> {
   }
 
   fun remove(key: Key): Maybe<Value>
+  fun removeWhere(selector: (Key, Value) => Bool) {
+    for entry in (this as Map<Key, Value>).entries() {
+      if selector(entry.first, entry.second) { remove(entry.first) }
```
MarcelGarus, line 55: Lucky us that we don't have a concurrent modification exception yet. 😅

packages/ast/src/lowering.candy
```diff
@@ -168,11 +186,1072 @@ fun lowerModifier(
...
+  }
+  if cst.child is CstClass {
+    return lowerClass(context, parentId, cst.cast<CstClass>(), previousNames)
+  }
+  if cst.child is CstFunction {
+    return lowerFunction(context, parentId, cst.cast<CstFunction>())
+  }
+  if cst.child is CstProperty {
+    return lowerProperty(context, parentId, cst.cast<CstProperty>(), previousNames)
+  }
```
MarcelGarus, line 230: If we would contain a stack of scopes containing the `parentId`s and `previousNames` into the `LoweringContext`, we could add a `Lowerable` trait (or something similar) implemented by `CstFile`, `CstModule`, `CstFunction` etc. so that we could do `cst.lower(context)` instead of doing manual dispatch based on the type.

packages/ast/src/lowering.candy
```diff
@@ -168,11 +186,1072 @@ fun lowerModifier(
...
+        ExpectedExpressionCompilerError(
+          context.file,
+          initializer.cast<CstExpressionBody | CstDelegationBody>(),
+        ),
+      )
+      None<AstExpression>()
+    } else {
+      Some<AstExpression>(lowerExpression(context, id, initializer.child.expression.unwrap()))
+    }
+  }
```
MarcelGarus: If we had a `flatten` method on `Maybe<Maybe<T>>`, we could have simply done:
```suggestion
  let initializer = cst.child.initializer.flatMap<AstExpression>({ initializer =>
    initializer.child.expression
      .map<Maybe<AstExpression>>({ lowerExpression(context, id, it) })
      .orElse({
        context.addError(
          ExpectedExpressionCompilerError(
            context.file,
            initializer.cast<CstExpressionBody | CstDelegationBody>(),
          ),
        )
      })
      .flatten()
  })
```
But we can still use map-orElse instead of the outer if.

packages/ast/src/lowering.candy
```diff
@@ -43,6 +46,17 @@ fun lowerFile(context: LoweringContext, cst: CstNode<CstFile>): AstFile {
...
+        context,
+        id,
+        node.cast<CstDeclaration>(),
+        previousDeclarationNames as Map<String, Bool>,
+      )
+      if result is Some<AstDeclaration> {
+        declarations.append(
+          result.unwrap() as AstModule | AstTrait | AstImpl | AstClass | AstFunction | AstProperty,
+        )
+      }
```
MarcelGarus: Maybe a `do` method on `Maybe` would be useful:
```suggestion
      result.do({
        declarations.append(
          it as AstModule | AstTrait | AstImpl | AstClass | AstFunction | AstProperty,
        )
      })
```

packages/ast/src/lowering.candy
```diff
@@ -168,11 +186,1072 @@ fun lowerModifier(
...
+  } else {
+    if previousNames.containsKey(name.unwrap().value) {
+      context.addError(
+        DuplicateDeclarationNameCompilerError(
+          context.file,
+          cst.cast<CstModule | CstTrait | CstClass | CstProperty>(),
+        ),
+      )
+    }
+  }
```
MarcelGarus: This has been a rollercoaster to read: First, a `name` is created, but it's only used for the `id` and then later overwritten by another name variable. Here's a simpler version (also, no `unwrap` is necessary):
```suggestion
  let id = AstDeclarationId(
    parentId,
    cst.child.name
      .map<String>({ it.child.identifier })
      .orElse({ "unnamed-module-{context.getUnnamedDeclarationIndex()}" })
  )
  let modifiers = lowerModifiers(context, id, cst.child.modifiers)
  let name = cst.child.name.map<AstIdentifier>({
    let identifierId = AstIdentifierId(id, 0)
    context.addIdMapping(it.id, identifierId)
    if previousNames.containsKey(it.child.identifier) {
      context.addError(
        DuplicateDeclarationNameCompilerError(
          context.file,
          cst.cast<CstModule | CstTrait | CstClass | CstProperty>(),
        ),
      )
    }
    AstIdentifier(identifierId, it.child.identifier)
  })
  if name is None {
    context.addError(MissingDeclarationNameCompilerError(context.file, cst.cast<CstDeclaration>()))
  }
```
Of course, this also applies to all cases where this code has been copied to.

packages/ast/src/lowering.candy
```diff
@@ -168,11 +186,1072 @@ fun lowerModifier(
...
+    context.addError(
+      ExpectedExpressionCompilerError(
+        context.file,
+        cst.cast<CstExpressionBody | CstDelegationBody>(),
+      ),
+    )
+    None<AstExpression>()
+  } else {
+    Some<AstExpression>(lowerExpression(context, parentId, cst.child.expression.unwrap()))
+  }
```
MarcelGarus:
```suggestion
  let expression = cst.child.expression
    .map<AstExpression>({ lowerExpression(context, parentId, it) })
  if expression is None {
    context.addError(
      ExpectedExpressionCompilerError(
        context.file,
        cst.cast<CstExpressionBody | CstDelegationBody>(),
      ),
    )
  }
```

packages/ast/src/lowering.candy
```diff
@@ -168,11 +186,1072 @@ fun lowerModifier(
...
+        node.cast<CstTypeParameter>(),
+        previousNames as Map<String, Bool>,
+      )
+      if lowered is Some<AstTypeParameter> {
+        let unwrapped = lowered.unwrap()
+        result.append(unwrapped)
+        if unwrapped.name is Some<AstIdentifier> {
+          previousNames.set(unwrapped.name.unwrap().value, true)
+        }
+      }
```
MarcelGarus: Could be rewritten if we had `Maybe.do`:
```suggestion
      lowered.do({
        result.append(it)
        it.name.do({ previousNames.set(it.value, true) })
      })
```

packages/ast/src/lowering.candy
```diff
@@ -307,6 +1392,8 @@ class LoweringContext {
       file,
       MutableMap.empty<CstNodeId, AstNodeId>(),
       0,
+      0,
+      0,
```
MarcelGarus: Ooof.
Maybe comments after the zeros?

MarcelGarus (changes requested):
Nice to see progress!

packages/ast/src/declarations.candy
JonasWanke: Yes. The parser just tries to parse any type. When lowering to the AST, we already check that this type is a named type or an intersection of named types, and flatten these into a list. Hence, the list doesn't contain arbitrary types, but only named types.

packages/ast/src/lowering.candy
JonasWanke: I'm not sure if a stack of scopes is worth it right now. To allow for type-safe access of the parent scope we'd need a type parameter in context declaring the parent node's type. Maybe we can postpone this refactor until we have a better compiler?

packages/ast/src/types.candy
JonasWanke: Not quite sure what you mean by that. Is it about the existence of `AstTypeArgumentId` or this change?

MarcelGarus: Yes, the existence of `AstTypeArgumentId`.

packages/ast/src/lowering.candy
MarcelGarus: Sure

MarcelGarus (approved):

PR #66, opened 2021-02-06
Add queries for CSTs and ASTs of files

JonasWanke:


MarcelGarus:
`someFile.cst().ast()` would be a nicer API and I think that's possible.
Edit: Not that easy after all.

MarcelGarus (approved):

PR #69, opened 2021-02-11
Dependency resolution

JonasWanke:
We now have queries to:

* get all dependencies of a project (including transitive ones)
* get the compilation order for a project (By compiling packages in the order of the returned list, packages are only compiled after all of their dependencies.)

Also, additions to the standard library and bugfixes as usual.


packages/compiler_utils/src/dependencies.candy
```diff
@@ -0,0 +1,81 @@
+use incremental
+use petit_parser
+use ..error
+use ..file
+use ..location
+use ..package
+
+impl Package {
+  public fun getAllDependencies(context: QueryContext<List<CompilerError>>): Set<Package> {
```
MarcelGarus, line 9: Can't we reduce `getAllDependencies` to `getAllPackagesInCompilationOrder`? As in, couldn't we implement `getAllDependencies` like this:
```kotlin
public fun getAllDependencies(context: QueryContext<List<CompilerError>>): Set<Package> {
  getAllPackagesInCompilationOrder(context).toSet()
}
```
Then, cycle checking would also need to be included in `getAllPackagesInCompilationOrder`.

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #79, opened 2021-02-22
Dart code generator beginnings

JonasWanke:
* <kbd>Core</kbd>: Add `Process.run(…)`, `ProcessResult`
* <kbd>CodeGeneratorDart</kbd>: generate `pubspec.yaml` and execute `dart pub get`

### DIFF
--- a/packages/code_generator_dart/src/compile.candy
+++ b/packages/code_generator_dart/src/compile.candy
@@ -1,0 +1,55 @@
+use compiler_utils
+use incremental
+
+public fun compile(context: QueryContext<List<CompilerError>>, package: Package): Unit {
+  impureQuery<Unit, List<CompilerError>>(context, "code_generator_dart.compile", package, {
+    let outputPath = package.getOutputPath(context, "Dart")
+    let outputDirectory = Directory(outputPath)
+    if outputDirectory.doesExist() {
+      outputDirectory.delete(recursive = true)
+      outputDirectory.create(recursive = true)
+    }
+
+    let files = compile_(context, package)
+    for entry in files.entries() {
+      let path = outputPath.append(entry.first)
+      let content = entry.second
+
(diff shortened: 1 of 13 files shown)


packages/code_generator_dart/src/compile.candy
```diff
@@ -0,0 +1,60 @@
...
+    let output = MutableMap.empty<Path, String>()
+
+    let pubspec = generatePubspec(context, package)
+    output.set(pubspec.first, pubspec.second)
+
+    // TODO(JonasWanke): generate code
+
+    Tuple(output, List.empty<CompilerError>())
+  })
+}
```
MarcelGarus, line 43: Maybe make this a method on `Package`? Usable as `let files = somePackage.compile(context)`.

packages/code_generator_dart/src/compile.candy
```diff
@@ -0,0 +1,60 @@
...
+  sdk: '>=2.12.0-0 <3.0.0'
+dependencies:
+  characters: ^1.1.0
+  meta: ^1.3.0
+dev_dependencies:
+  test: ^1.16.3"
+
+    Tuple(Tuple(path, content), List.empty<CompilerError>())
+  })
+}
```
MarcelGarus, line 60: Same here: `myPackage.generatePubspec(context)` would be cleaner.

MarcelGarus (approved):

packages/code_generator_dart/src/compile.candy
JonasWanke: We'll have to refactor this when our compiler supports impls for types from other packages.

packages/code_generator_dart/src/compile.candy
JonasWanke: See the other comment.

PR #80, opened 2021-02-23
DartCode improvements

JonasWanke:

### DIFF
--- a/packages/dart_code/src/nodes.candy
+++ b/packages/dart_code/src/nodes.candy
@@ -1,4 +1,11 @@
+public let dartPubspecFileName = "pubspec.yaml"
+public let dartLibPath = Path.parse("lib")
+public let dartSourcePath = dartLibPath.child("src")
+public let dartTestPath = Path.parse("test")
+public let dartFileExtension = "dart"
+
 class DartCompilationUnit {
+
   let directives: List<DartDirective>
   let declarations: List<DartDeclaration>
 }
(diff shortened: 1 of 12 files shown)


packages/dart_code/src/nodes.candy
```diff
@@ -156,13 +190,44 @@ impl DartBlock: DartBody
...
+static fun dartIterator(entry: DartType): DartType {
+  DartNamedType(DartIdentifier("Iterator", Some<String>(dartCoreUrl)), List.of1<DartType>(entry))
+}
+static fun dartList(entry: DartType): DartType {
+  DartNamedType(DartIdentifier("List", Some<String>(dartCoreUrl)), List.of1<DartType>(entry))
+}
+static fun dartMap(key: DartType, value: DartType): DartType {
+  DartNamedType(DartIdentifier("Map", Some<String>(dartCoreUrl)), List.of2<DartType>(key, value))
+}
+static let dartNull = DartNamedType(DartIdentifier("Null", Some<String>(dartCoreUrl)), List.empty<DartType>())
```
MarcelGarus, line 216: We should make sure this doesn't get confused with the `null` constant (for which we should probably also have a constant here.

MarcelGarus (approved):

packages/dart_code/src/nodes.candy
JonasWanke: For `null`, there's `DartNullLiteral()` already. Should there be a constant for that too? I'm not sure that's worth it, unless we make `DartNullLiteral` private.

Regarding the type constant: Should I rename it to `dartNullType`? That would differentiate it from other type constants, but it's clearer that it's a type. (This becomes irrelevant as soon as we can have static properties in traits, as we can move these to `DartType`.)

MarcelGarus: Ah, didn't spot the `DartNullLiteral`. Then I'd leave this as-is, keep it consistent.

MarcelGarus:
I believe this can be merged now?

PR #85, opened 2021-03-24
Update readme

MarcelGarus:


README.md
```diff
@@ -1,12 +1,26 @@
 # 🍭 Candy
 
-A sweet programming language, mainly inspired by Kotlin, Rust and Dart.
+A sweet programming language, mainly inspired by Kotlin, Rust, and Dart.
```
JonasWanke: Maybe add Elixir?

README.md
```diff
@@ -47,63 +61,69 @@ A sweet programming language, mainly inspired by Kotlin, Rust and Dart.
...
   }
   ```
-  Note we also want to add pattern matching on the language-level, but we didn't think about how to do that yet, because other features were more urgent.
+  Note we also want to add pattern matching on the language-level, but we didn't decide on a final design yet, because other features were more urgent.
 * **Be modular.**
   All elements are private by default. The `public` modifier makes functions, properties and types available if the module gets imported.
-  The module system itself is very much inspired by Rust – folders, files, and explicit `module` declarations can all be used interchangably. Also, defining types creates implicit modules.  
-  Our imports are called *use-lines*, because they look like this:
+  Rust very much inspires the module system itself – you can use folders, files, and explicit `module` declarations interchangeably.
+  Also, defining types creates implicit modules.
```
JonasWanke: I think this only applies to traits and not our `type` declaration.

JonasWanke (changes requested):

JonasWanke (approved):

PR #87, opened 2021-03-24
Migrate HIR

MarcelGarus:
This PR migrates the majority of the HIR to the new philosophy.
It also fixes some bugs.


new-packages/Core/src/Bool.candy
```diff
@@ -0,0 +1,17 @@
+use ..Maybe
+## use ..operators
```
JonasWanke:
```suggestion
## use ..Operators
```

new-packages/Core/src/Bool.candy
```diff
@@ -0,0 +1,17 @@
...
+
+public type Bool = True | False
+
+fun true(): Bool { Bool.True }
+fun false(): Bool { Bool.False }
```
JonasWanke:
```suggestion
public fun true(): Bool { Bool.True }
public fun false(): Bool { Bool.False }
```

packages/core/src/collections/map/module.candy
```diff
@@ -71,6 +71,17 @@ public trait Map<Key: Equals & Hash, Value> {
     }
     map
   }
+
+  fun unsafeEquals(other: Map<Key, Value>): Bool {
+    keys().toSet().unsafeEquals(other.keys().toSet())
+      && keys().map<Bool>({ (get(it) as Equals) == (other.get(it) as Equals) }).all({ it })
```
JonasWanke:
```suggestion
      && keys().all({ (get(it) as Equals) == (other.get(it) as Equals) })
```

packages/hir/src/example.candy
```diff
@@ -40,7 +40,9 @@ fun useLinesExample(context: QueryContext<List<CompilerError>>) {
 
 fun resolvingExample(context: QueryContext<List<CompilerError>>) {
   print("Resolving Example")
-  let file = FancyFile(Package.core(context), Path.parse("src/module.candy"))
+  let file = FancyFile(Package.core(context), Path.parse("src/.candy"))
+  let ast = fileToHirModule(context, file).innerModule("Primitives").astDeclarations(context)
+  print("Ast primitives: {ast}")
```
JonasWanke:
```suggestion
  print("AST primitives: {ast}")
```

packages/hir/src/lowering/declarations.candy
```diff
@@ -21,7 +23,7 @@ fun fileToHirModule(context: QueryContext<List<CompilerError>>, file: FancyFile)
 
   query<HirModule, List<CompilerError>>(context, "HirModule.fromFile", (file as Equals & Hash), {
     mut let path = file.path
-    if path.baseNameWithoutExtension() == "module" {
+    if path.baseNameWithoutExtension() == ".candy" {
```
JonasWanke, line 26:
```suggestion
    if path.baseName() == ".candy" {
```

packages/hir/src/lowering/declarations.candy
```diff
@@ -342,12 +340,12 @@ impl HirClass: HasAst<AstClass> & HasDeclarations {
 
   fun astDeclarations(
```
JonasWanke: Is this function still necessary, or can we remove this function and just return an empty list from `typeUnsafeAstDeclarations(…)` below?

packages/hir/src/lowering/types.candy
```diff
@@ -43,24 +43,39 @@ fun astTypeToHirType(
             .expand<HirDeclaration>({
               namePart.resolveIdentifier(context, it) as Iterable<HirDeclaration>
             })
-            .toSet() as Set<HirClass | HirTrait | HirProperty | HirFunction | HirModule>)
+            .toSet() as Set<HirType | HirTrait | HirFunction | HirModule>)
         }
 
         // There should only be one candidate left.
```
JonasWanke: Now that I think about it, one is the maximum, but this could also return no candidates at all, right? If so, we should return a `HirErrorType`.

packages/hir/src/lowering/types.candy
```diff
@@ -43,24 +43,39 @@ fun astTypeToHirType(
...
+          if types.any({ it is None }) {
+            HirErrorType()
+          } else {
+            HirNamedType(
+              declaration as HirTrait | HirType,
+              types.map<HirInlineType>({ it.unwrap() }).toList(),
+            )
+          },
+          List.empty<CompilerError>(),
+        )
```
JonasWanke: We could return the actual named type and just mark specific type arguments as errors. (And with the next PR, the type arguments's `type` is no longer a `Maybe`.

```suggestion
        let types = ((astType as AstNamedType).typeArguments as Iterable<AstTypeArgument>)
          .map<HirInlineType>({
            it.type
              .map<HirInlineType>({ astInlineTypeToHirInlineType(context, it, scope) })
              .orElse({ HirErrorType() })
          })
          .toList()
        return Tuple(
          HirNamedType(declaration as HirTrait | HirType, types),
          List.empty<CompilerError>(),
        )
```

packages/hir/src/types.candy
```diff
@@ -1,53 +1,119 @@
...
+  fun hash<T>(hasher: Hasher<T>) { types.unsafeHash<T>(hasher) }
 }
-impl HirFunctionType: HirType
 
 
-public class HirUnionType {
-  public let types: List<HirType>
-  // TODO(JonasWanke): Use a `Set<HirType>` when we have sets that keep insertion order.
+public class HirEnumType {
+  public let variants: Map<String, HirInlineType>
```
JonasWanke: Enum variants don't have to contain a value.

```suggestion
  public let variants: Map<String, Maybe<HirInlineType>>
```

JonasWanke: Maybe use a list to keep the order, e.g., when displaying this type in a tooltip?

packages/hir/src/types.candy
```diff
@@ -1,53 +1,119 @@
...
 }
-impl HirTupleType: HirType
 
 
-public class HirFunctionType {
-  public let receiverType: Maybe<HirType>
-  public let parameterTypes: List<HirType>
-  public let returnType: HirType
+public class HirNamedTupleType {
+  public let types: Map<String, HirInlineType>
```
JonasWanke: Maybe use a list to keep the order, e.g., when displaying this type in a tooltip? Also, I think we wanted to allow default values in the future, which requires ordering when using them.

packages/hir/src/lowering/types.candy
```diff
@@ -69,37 +84,120 @@ fun astTypeToHirType(
...
+          } else {
+            HirEnumType(
+              (loweredVariants as Iterable<Maybe<(String, HirInlineType)>>)
+                .map<(String, HirInlineType)>({ it.unwrap() })
+                .unsafeToMap<String, HirInlineType>(),
+            )
+          },
+          List.empty<CompilerError>(),
+        )
+      }
```
JonasWanke: Enum variants don't have to contain a value.

```suggestion
      if (astType is AstEnumType) {
        let loweredVariants = ((astType as AstEnumType).variants as Iterable<AstEnumTypeVariant>)
          .map<(String, Maybe<HirInlineType>)>({
            let valueType = it.valueType.map<(String, HirInlineType)>({ valueType =>
              astInlineTypeToHirInlineType(context, valueType, scope)
            })
            Tuple(it.name.value, valueType)
          })
          .toList()
        return Tuple(HirEnumType(loweredVariants, List.empty<CompilerError>())
      }
```

JonasWanke (changes requested):

packages/hir/src/lowering/types.candy
```diff
@@ -46,23 +46,36 @@ fun astInlineTypeToHirInlineType(
...
+            List.of1<CompilerError>(CouldNotResolveTypeCompilerError(astType as AstNamedType)),
+          )
+        }
+        if (candidates as Iterable).length() > 1 {
+          return Tuple(
+            HirErrorType(),
+            List.of1<CompilerError>(
+              //       // This cast from `Set<HirType | HirTrait | HirFunction | HirModule>` to
+              //       // `Set<HirType | HirTrait | HirModule>`succeeds because the name is uppercase, so no
+              //       // functions get resolved.
```
JonasWanke:
```suggestion
              // This cast from `Set<HirType | HirTrait | HirFunction | HirModule>` to
              // `Set<HirType | HirTrait | HirModule>` succeeds because the name is uppercase, so no
              // functions get resolved.
```

packages/hir/src/lowering/types.candy
```diff
@@ -201,3 +210,35 @@ impl ConflictingImplsCompilerError: CompilerError {
...
+    todo("Implement AmbiguousTypeCompilerError.location") // TODO(marcelgarus)
+  }
+  public fun title(): String { "Ambiguous type. Candidates are {candidates}." }
+  public fun description(): String {
+    /// TODO(marcelgarus): add a description for this error
+    ""
+  }
+}
+
+public class CouldNotResolveTypeCompilerError {
```
JonasWanke: `UnknownTypeCompilerError` would be shorter, but the current name is fine too.

JonasWanke (approved):
Just one minor formatting issue and a name suggestion

PR #101, opened 2021-04-06
Start implementing type solving

MarcelGarus:
This adds the fundamentals for type solving.

For example, given these type definitions:
```
type Foo = Unit
trait Bar {}
impl Foo: Bar {}
```
We can now lower the impl into this `SolverRule`: `Foo(Bar)`

Next steps:
- Make the lowering more robust. Specifically, an impl like `impl[T] Foo[T]: Bar` can't get lowered yet, because `Foo[T]` gets lowered to an invalid HIR type (`Foo[HirErrorType]`).
- Support getting _all_ impls.
- Turn all impls into `SolverRule`s.
- Check that no two impls for the same trait can be unified. This disallows `impl[T] Foo[Int, T]: Bar` and `impl[T] Foo[T, String]: Bar`, because the implementation of `Bar` for `Foo[Int, String]` would be ambiguous.
- Put all those `SolverRule`s into a solver `Environment`. This represents the type system and can be queried against.
- Add queries for checking whether a type implements a trait and if a type contains a method with a given signature.

### DIFF
--- a/new-packages/Core/src/Bool.candy
+++ b/new-packages/Core/src/Bool.candy
@@ -1,5 +1,5 @@
 use ..Maybe
-## use ..Operators
+use ..Operators
 
 public type Bool = True | False
 
--- a/new-packages/Playground/src/Types.candy
+++ b/new-packages/Playground/src/Types.candy
@@ -1,0 +1,5 @@
+type Foo = Unit
+
+trait Bar {}
+
+impl Foo: Bar {}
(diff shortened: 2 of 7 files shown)


JonasWanke (approved):

MarcelGarus:
I still use the `String.resolveIdentifier` code instead of the top-level method, so I'm gonna manually merge the main branch before merging.
Edit: Never mind, it was all migrated already

PR #105, opened 2021-04-08
Honor use-lines when resolving identifiers

JonasWanke:

### DIFF
--- a/packages/hir/src/lowering/types.candy
+++ b/packages/hir/src/lowering/types.candy
@@ -53,7 +53,6 @@ fun astInlineTypeToHirInlineType(
         }
         let candidates = candidates.items()
           .where({ !(it is HirModule) })
-          .cast<HirType | HirTrait | HirFunction | HirParameterType>()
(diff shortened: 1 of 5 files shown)


packages/hir/src/lowering/types.candy
```diff
@@ -41,12 +41,24 @@ fun astInlineTypeToHirInlineType(
...
+              } else {
+                List.empty<HirModule | HirType | HirTrait | HirFunction | HirParameterType>()
+              }
             })
-            .toSet() as Set<HirType | HirTrait | HirFunction | HirModule>)
+            .toSet()
         }
+        let candidates = candidates.items()
+          .where({ !(it is HirModule) })
+          .cast<HirType | HirTrait | HirFunction | HirParameterType>()
```
MarcelGarus: We don't need this line because it's directly followed by another cast.
```suggestion
```

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #123, opened 2021-05-20
Refactor hash and do some small fixes

MarcelGarus:


new-packages/Core/src/Io/Process.candy
```diff
@@ -3,14 +3,11 @@ use ...Int
...
 
-public builtin type Process
-impl Process {
-  public static builtin fun run(
-    executable: Path,
-    arguments: List[String],
-    workingDirectory: Path,
-  ): ProcessResult
-}
+public static builtin fun runProcess(
```
JonasWanke:
```suggestion
public builtin fun runProcess(
```

new-packages/Core/src/Hash.candy
```diff
@@ -2,37 +2,34 @@ use ..Collections
...
-
-  fun hashCode(): Int { Hasher.default().write(this).finish() }
+  # A type that can be hashed.
+
+  fun hash[H: Hasher](): H
+  # Uses `H`'s static `primitive` and `combine` functions to hash primitive bytes into `H`s and
+  # combine multiple `H`s into one single `H`.
+  #
+  # The `H` can be used to customize how good and fast the hash is. For example, for simple
+  # microcontroller programms, you might want to hash into a `UInt16` efficiently, but for a big
```
JonasWanke:
```suggestion
  # microcontroller programs, you might want to hash into a `UInt16` efficiently, but for a big
```

new-packages/Core/src/Hash.candy
```diff
@@ -2,37 +2,34 @@ use ..Collections
...
-  }
-  fun write(value: List[Int]): Hasher[Int] {
-    value.foldLeft[Hasher[Int]](this, { result, item -> result.write(item) })
-  }
-
-  fun finish(): Int { this.entropy }
+type IntHasher = Int
+impl IntHasher: Hasher {
+  static fun primitive(value: Int): This { IntHasher(value) }
+  static fun combine(values: Iterate[This]): This { values.map({ it.value }).sum() }
```
JonasWanke:
```suggestion
  static fun combine(values: Iterate[This]): This { values.sumBy({ it.value }) }
```

new-packages/Core/src/Result.candy
```diff
@@ -19,3 +19,15 @@ impl[Success, ErrorType] Result[Success, ErrorType] {
...
+impl[Success: Equals, ErrorType: Equals] Result[Success, ErrorType]: Equals {
+  fun equals(other: Result[Success, ErrorType]): Bool {
+    this.match[Bool](
+      ok = { thisSuccess -> other.match[Bool](ok = { thisSuccess == it }, error = { false() }) },
+      error = { thisError -> other.match[Bool](ok = { false() }, error = { thisError == it }) },
+    )
+  }
+}
+impl[Success: Hash, ErrorType: Hash] Result[Success, ErrorType]: Hash {
+  fun hash[H: Hasher](): H { this.match[H](ok = { it.hash[H]() }, error = { it.hash[H]() }) }
```
JonasWanke: If `Result Success` and `Result Error` contain the same value, their hashes would be identical. Maybe we should also `combine` this with a constant representing the case?

new-packages/Core/src/String.candy
```diff
@@ -20,11 +20,21 @@ impl String {
   public builtin fun trim(): String
 }
 
-builtin impl String: Equals & Comparable
+impl String: Equals {
+  ## TODO(marcelgarus): Implement `Compare` instead, which automatically implements `Equals`.
+
+  fun equals(other: String): Bool {
+    this.length() == other.length()
+      & this.characters().zip(other.characters()).all({ it.first == it.second })
```
JonasWanke: Unfortunately, that won't work – the items returned by `characters()` are `String`s themselves, leading to infinite recursion. Until we handle Unicode in Candy directly, we'll probably have to make this `builtin`.

new-packages/Core/src/String.candy
```diff
@@ -20,11 +20,21 @@ impl String {
...
+}
+
 impl String: Hash {
-  fun hash[Result](hasher: Hasher[Result]): Hasher[Result] {
-    hasher.write(this.length())
-    hasher.write(this.characters().map[Int]({ 0 }).toList())
+  fun hash[H: Hasher](): H {
+    H.combine(List.of(
+      H.primitive(this.length),
+      H.combine(this.characters().map({ it.hash[H]() })),
```
JonasWanke: Same as above: This will lead to infinite recursion.

JonasWanke (changes requested):

JonasWanke (approved):

PR #126, opened 2021-05-29
Core: Ranges & Progressions

JonasWanke:
Some things to note:

* I didn't add an `UnboundedIntProgression` yet, which would probably have a `step` and some `alignment` (to be able to offer a `contains` method).
* The range syntax used in the doc comments of ranges is not final. I'm not a fan of `a=..b` being the syntax for the probably most common range with an inclusive start and exclusive end.
* I didn't add a `Step` trait yet as we decided to only offer `IntProgression`s for now.

### DIFF
--- a/new-packages/Core/src/Ranges/Progressions.candy
+++ b/new-packages/Core/src/Ranges/Progressions.candy
@@ -1,0 +1,114 @@
+use ...Bool
+use ...Collections
+use ...Int
+use ...Maybe
+use ...Operators
+
+public trait IntProgression: Equals {
+  fun startBound(): Maybe[Int]
+  fun endBound(): Maybe[Int]
+  fun stepSize(): Int
+
+  fun contains(item: Int): Bool {
+    let startMatches = this.startBound()
+      .map[Bool]({ it <= item & (item - it) % this.stepSize() == 0 })
+      .else({ true() })
+    let endMatches = this.endBound()
+      .map[Bool]({ item <= it & (it - item) % this.stepSize() == 0 })
+      .else({ true() })
+    startMatches & endMatches
+  }
+}
+
+
+public type BoundedIntProgression = (start: Int, end: Int, step: Int)
+# A `Progression` where `this.start` is inclusive and `this.end` is exclusive.
+#
+# Invariants:
+# * `this step != 0`
+# * `(this step > 0) then { this end > this start } else { this end < this start }`
+# * `(this end - 1 - this start) % this step == 0`
+
+impl BoundedIntProgression: IntProgression {
+  fun startBound(): Maybe[Int] { Maybe[Int].Some(this.start) }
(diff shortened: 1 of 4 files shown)


new-packages/Core/src/Ranges/Progressions.candy
```diff
@@ -0,0 +1,114 @@
...
+  }
+}
+
+
+public type BoundedIntProgression = (start: Int, end: Int, step: Int)
+# A `Progression` where `this.start` is inclusive and `this.end` is exclusive.
+#
+# Invariants:
+# * `this step != 0`
+# * `(this step > 0) then { this end > this start } else { this end < this start }`
```
MarcelGarus: How is an empty progression represented so that these two invariants are true?

new-packages/Core/src/Ranges/Progressions.candy
```diff
@@ -0,0 +1,114 @@
...
+# * `(this end - 1 - this start) % this step == 0`
```
MarcelGarus: Is this what we want? Should `BoundedIntProgression(step = 3, start = 0, end = 4)` be valid?

new-packages/Core/src/Ranges/Progressions.candy
```diff
@@ -0,0 +1,114 @@
...
+    })
+  }
+
+  ## TODO(JonasWanke): Add a constant-time `contains(item)` when we have something like
+  ## where-clauses for generic types and `iterate.contains` can be declared directly inside
+  ## `Iterate`. Same for the progressions below.
+
+  fun length(): Int {
+    if (this.start == this.end, { 0 })
+      .else({ this.end - 1 - this.start) // this.step + 1 })
```
MarcelGarus, line 60:
```suggestion
      .else({ this.end - this.start) // this.step })
```
At least when the invariant above is fixed. And the `//` is basically exact division every time

MarcelGarus (changes requested):

new-packages/Core/src/Ranges/Progressions.candy
JonasWanke: Fixed the invariant: `this end` is equal to `this start` to represent an empty progression.

new-packages/Core/src/Ranges/Progressions.candy
JonasWanke: Yes, that should be what we discussed before. Or do you think we should encode progressions differently?

MarcelGarus: I somehow thought we had the end as an exclusive range, but it's `step` larger than the last yielded value. So, for the progression `0, 3, 6, 9` it would be `step = 3, start = 0, end = 12`. So, basically, `(end - start) % step == 0`, and `(end - start) / step` is the number of items in the progression.

MarcelGarus (approved):

PR #131, opened 2021-07-17
Rename tuple to positional tuple

MarcelGarus:


new-packages/Core/src/Tuples.candy
```diff
@@ -6,5 +6,6 @@
 # parenthesis to create them on-the-fly:
 #
 # ```
-# let a: (String, Int) = ("foo", 42)
+# let a: (String, Int) = ("foo", 42) # Positional tuple
+# let b: (foo: String, bar: Int) = (foo: "foo", bar: 42) # Named tuple
```
JonasWanke, line 8: Line three is outdated: named tuples give their fields names

packages/hir/src/declarations.candy
```diff
@@ -36,7 +36,7 @@ public trait HirModule: HirDeclaration & Equals & Hash {
   static fun syntheticNamedTuples(context: QueryContext<List<CompilerError>>): HirModule {
     synthetic(context).innerModule("NamedTuples")
   }
-  static fun syntheticTuples(context: QueryContext<List<CompilerError>>): HirModule {
+  static fun syntheticPositionalTuples(context: QueryContext<List<CompilerError>>): HirModule {
```
JonasWanke: This function can be removed

JonasWanke (changes requested):

JonasWanke (approved):

PR #150, opened 2022-02-08
Differentiate values and functions in assignments

JonasWanke:
image


new-new-packages/compiler/src/analyzer/mod.rs
```diff
@@ -1,58 +1,91 @@
...
+            Some(Ok(value)) => Some(AnalyzerReport::ValueOfExpression {
+                id: id.to_owned(),
+                value,
+            }),
+            Some(Err(value)) => Some(AnalyzerReport::ExpressionPanics {
+                id: id.to_owned(),
+                value,
+            }),
+            None => None,
+        })
```
MarcelGarus: We could first do a `filter_map(|it| it)` to filter out `None`s followed by a normal `map`. That would simplify the code in the `match`, not sure if it's more beautiful overall.

MarcelGarus (approved):

PR #157, opened 2022-03-04
Builtin struct methods, Maybe in core, and refactorings

JonasWanke:


MarcelGarus (commented):
Nice! Makes the code much cleaner.

MarcelGarus (approved):

PR #161, opened 2022-03-28
Add use function

JonasWanke:


new-new-packages/compiler/src/discover/run.rs
```diff
@@ -17,160 +17,164 @@ use super::{
...
+    //     }
+    // } else {
+    //     let (hir, _) = db.hir(function.input.to_owned()).unwrap();
+    //     hir.as_ref().to_owned()
+    // };
+    // match lambda_parent.identifiers.get(&function) {
+    //     Some(function_name) => log::trace!("Calling function `{}`", &function_name),
+    //     // TODO: Inline lambdas don't have names, but for some other lambdas the name is not found.
+    //     None => log::trace!("Calling function {}", function),
+    // };
```
MarcelGarus: What's up with this?

MarcelGarus (approved):

new-new-packages/compiler/src/discover/run.rs
JonasWanke: This was logging for every single function call, which quickly adds up when a whole file is invalidated at once, and, therefore, also seemingly slowed the VS Code Extension down a bit. I'm not sure how we want to proceed with this in the future – debug logging of the called functions might still be useful, especially when running the compiler from the console, but in VS Code, it's too much. Maybe conditionally run this only from the CLI?

PR #173, opened 2022-04-12
New parser

MarcelGarus:
This PR contains changes in three areas (and contains still only about 6000 lines of changes 😅):

- **New parser:** We don't use `nom` anymore. Instead, we use a completely new, handwritten parser that allows for better error recovery. In lots of subtle places, it has a different parsing behavior than the existing parser. For example, indentation is now more important than parentheses and invalid characters in the code are handled more gracefully.  
  The parser also has a new stage, the _raw concrete syntax tree_ (RCST for short), which sits before the CST stage and doesn't contain offsets or IDs. That's really useful for debugging and writing tests.

- **Debugging experience:** Additionally to the `lsp` and `run` subcommands, there's now a new `build` subcommand that could later be used to compile the code to some target format instead of running it immediately. There are also two options for the `build` command:
  - The `--watch` option watches the filesystem for changes and automatically rebuilds if you edit the input file.
  - The `--debug` option generates debug output files next to the input file. Those contain the RCST, CST, AST, and HIR, as well as all mappings between IDs.

- **Flow of errors:** The errors are not immediately turned into `String`s, but used as enums instead. That makes it possible to display errors differently depending on the medium and also works better with the tooling, such as finding all places where a particular error is generated.  
  Errors are also not a by-product of compiling to the next stage anymore. Instead, the RCST, CST, AST, and HIR trees directly contain all errors, including those of previous compilation stages.
  The compiler also doesn't abort anymore if an error is found. The goal is to allow you to run even a file with syntax errors (although the program will panic if it encounters an unparsable section). Here's a file that reports RCST, AST, and HIR errors simultaneously:<br>
  image

In-person, we also talked about implementing hierarchical CST IDs, but those will come in a separate PR.


new-new-packages/compiler/src/compiler/hir.rs
```diff
@@ -173,25 +190,26 @@ impl fmt::Display for Expression {
                     "struct [\n{}\n]",
                     entries
                         .iter()
-                        .map(|(id, value)| format!("  {}: {}", id, value))
-                        .join(",\n"),
+                        .map(|(key, value)| format!("  {}: {},", key, value))
+                        .join("\n"),
```
JonasWanke, line 194: With this change, we no longer add trailing commas. Is this intentional?

new-new-packages/compiler/src/compiler/hir.rs
```diff
@@ -203,15 +221,41 @@ impl fmt::Display for Expression {
...
-        write!(f, "{} parameters\n", self.parameters.len())?;
+        write!(
+            f,
+            "{} {} ->\n",
+            self.parameters.len(),
+            if self.parameters.len() == 1 {
+                "parameter"
+            } else {
+                "parameters"
+            }
```
JonasWanke: Can we write the actual parameter names instead?

new-new-packages/compiler/src/compiler/ast.rs
```diff
@@ -88,3 +96,144 @@ impl Deref for AstString {
...
+
+#[derive(Debug, PartialEq, Eq, Clone, Hash)]
+pub enum AstError {
+    UnexpectedPunctuation,
+    TextWithoutClosingQuote,
+    ParenthesizedWithoutClosingParenthesis,
+    CallOfANonIdentifier,
+    StructWithNonStructField,
+    StructWithoutClosingBrace,
+    ColonMissingAfterStructKey,
```
JonasWanke: Some errors indicating missing elements are called "WithoutFoo", one "FooMissing", and one "ExpectedFoo". Can we make this more uniform, e.g., by renaming this to `StructKeyWithoutColon`?

new-new-packages/compiler/src/compiler/ast.rs
```diff
@@ -88,3 +96,144 @@ impl Deref for AstString {
...
+                }
+            }
+            AstKind::Lambda(lambda) => lambda.body.collect_errors(errors),
+            AstKind::Call(call) => call.arguments.collect_errors(errors),
+            AstKind::Assignment(assignment) => assignment.body.collect_errors(errors),
+            AstKind::Error {
+                child,
+                errors: recovered_errors,
+            } => {
+                errors.append(&mut recovered_errors.clone());
```
JonasWanke: I think this might be more efficient:

```suggestion
                errors.extend(recovered_errors.into_iter());
```

(Or maybe just `iter`)

new-new-packages/compiler/src/compiler/ast.rs
```diff
@@ -88,3 +96,144 @@ impl Deref for AstString {
...
+                write!(
+                    f,
+                    "struct [\n{}\n]",
+                    struct_
+                        .fields
+                        .iter()
+                        .map(|(key, value)| format!("{}: {},", key, value))
+                        .join("\n")
+                        .lines()
+                        .map(|line| format!("  {}", line))
```
JonasWanke, line 169:
```suggestion
                        .map(|(key, value)| format!("  {}: {},", key, value))
```

new-new-packages/compiler/src/compiler/ast.rs
```diff
@@ -88,3 +96,144 @@ impl Deref for AstString {
...
+                        .iter()
+                        .map(|it| format!("{}", it))
+                        .join(" "),
+                    lambda
+                        .body
+                        .iter()
+                        .map(|it| format!("{}", it))
+                        .join("\n")
+                        .lines()
+                        .map(|line| format!("  {}", line))
```
JonasWanke, line 188:
```suggestion
                        .map(|it| format!("  {}", it))
```

new-new-packages/compiler/src/compiler/ast.rs
```diff
@@ -88,3 +96,144 @@ impl Deref for AstString {
...
+                )
+            }
+            AstKind::Call(call) => {
+                write!(
+                    f,
+                    "call {} with these arguments: {}",
+                    call.name,
+                    call.arguments
+                        .iter()
+                        .map(|argument| format!("{}", argument))
```
JonasWanke: Maybe add some indentation here?

new-new-packages/compiler/src/compiler/ast.rs
```diff
@@ -88,3 +96,144 @@ impl Deref for AstString {
...
+                    f,
+                    "assignment: {} =\n{}",
+                    assignment.name,
+                    assignment
+                        .body
+                        .iter()
+                        .map(|it| format!("{}", it))
+                        .join("\n")
+                        .lines()
+                        .map(|line| format!("  {}", line))
```
JonasWanke, line 214:
```suggestion
                        .map(|it| format!("  {}", it))
```

new-new-packages/compiler/src/compiler/ast.rs
```diff
@@ -88,3 +96,144 @@ impl Deref for AstString {
...
+                    write!(f, "\n  fallback: {}", child)?;
+                }
+                Ok(())
+            }
+        }
+    }
+}
+impl Display for AstString {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{}@{}", self.id, self.value)
```
JonasWanke:
```suggestion
        write!(f, "{}@\"{}\"", self.id, self.value)
```

new-new-packages/compiler/src/compiler/rcst.rs
```diff
@@ -0,0 +1,357 @@
...
+        error: RcstError,
+    },
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Hash)]
+pub enum RcstError {
+    IdentifierContainsNonAlphanumericAscii,
+    SymbolContainsNonAlphanumericAscii,
+    IntContainsNonDigits,
+    TextDoesNotEndUntilInputEnds,
```
JonasWanke: To unify this with `StructNotClosed` and others:

```suggestion
    TextNotClosed,
```

new-new-packages/compiler/src/compiler/rcst.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+}
+
+pub trait IsMultiline {
+    fn is_multiline(&self) -> bool;
+}
+
+impl IsMultiline for Rcst {
+    fn is_multiline(&self) -> bool {
+        log::info!("Is {:?} multiline?", self);
```
JonasWanke: Is this still necessary? If so, it can probably be downgraded to `debug` or `trace`

new-new-packages/compiler/src/compiler/rcst.rs
```diff
@@ -0,0 +1,357 @@
...
+            Rcst::Whitespace(_) => false,
+            Rcst::Newline(_) => true,
+            Rcst::Comment { .. } => false,
+            Rcst::TrailingWhitespace { child, whitespace } => {
+                log::info!("Is child multiline?");
+                let c = child.is_multiline();
+                log::info!("Is whitespace multiline?");
+                let w = whitespace.is_multiline();
+                log::info!("Combining");
+                c || w
```
JonasWanke: Is the logging necessary?

new-new-packages/compiler/src/compiler/rcst.rs
```diff
@@ -0,0 +1,357 @@
...
+                    || closing_parenthesis.is_multiline()
+            }
+            Rcst::Call { name, arguments } => name.is_multiline() || arguments.is_multiline(),
+            Rcst::Struct {
+                opening_bracket,
+                fields,
+                closing_bracket,
+            } => {
+                opening_bracket.is_multiline()
+                    || fields.iter().any(|field| field.is_multiline())
```
JonasWanke:
```suggestion
                    || fields.is_multiline()
```

new-new-packages/compiler/src/compiler/cst.rs
```diff
@@ -478,24 +467,30 @@ impl TreeWithIds for Cst {
...
+            CstKind::TrailingWhitespace { child, .. } => child.find_by_offset(offset),
             CstKind::Identifier { .. } => None,
             CstKind::Symbol { .. } => None,
-            CstKind::LeadingWhitespace { child, .. } => child.find_by_offset(offset),
-            CstKind::LeadingComment { child, .. } => child.find_by_offset(offset),
-            CstKind::TrailingWhitespace { child, .. } => child.find_by_offset(offset),
-            CstKind::TrailingComment { child, .. } => child.find_by_offset(offset),
+            CstKind::Int { .. } => None,
+            CstKind::Text { .. } => None,
+            CstKind::TextPart(_) => todo!(),
```
JonasWanke: TODOs

new-new-packages/compiler/src/database.rs
```diff
@@ -28,7 +28,8 @@ use crate::{
     LspPositionConversionStorage,
     ReferencesDbStorage,
     SemanticTokenDbStorage,
-    StringToCstStorage
+    StringToRcstStorage,
+    RcstToCstStorage
```
JonasWanke: Can you please insert this alphabetically?

new-new-packages/vscode_extension/src/extension.ts
```diff
@@ -46,6 +46,30 @@ async function spawnServer(): Promise<StreamInfo> {
 
   process.stderr.on('data', (data) => console.error(data.toString()));
 
+  process.addListener('close', (exitCode) => {
+    if (exitCode == 101) {
+      console.error('LSP server was closed with a panic.');
+    } else {
+      console.error('LSP server was closed with code ' + exitCode + '.');
```
JonasWanke:
```suggestion
      console.error(`LSP server was closed with code ${exitCode}.`);
```

new-new-packages/vscode_extension/src/extension.ts
```diff
@@ -46,6 +46,30 @@ async function spawnServer(): Promise<StreamInfo> {
...
+    console.error('LSP server disconnected.');
+  });
+  process.addListener('error', (event) => {
+    console.error('LSP server had an error: ' + event);
+  });
+  process.addListener('exit', (exitCode) => {
+    if (exitCode == 101) {
+      console.error('LSP server panicked.');
+    } else {
+      console.error('LSP server exited with exit code ' + exitCode + '.');
```
JonasWanke:
```suggestion
      console.error(`LSP server exited with exit code ${exitCode}.`);
```

JonasWanke:
```suggestion
      console.error(`LSP server exited with exit code ${exitCode}.`);
```

new-new-packages/vscode_extension/src/extension.ts
```diff
@@ -46,6 +46,30 @@ async function spawnServer(): Promise<StreamInfo> {
...
+      console.error('LSP server was closed with a panic.');
+    } else {
+      console.error('LSP server was closed with code ' + exitCode + '.');
+    }
+  });
+  process.addListener('disconnect', () => {
+    console.error('LSP server disconnected.');
+  });
+  process.addListener('error', (event) => {
+    console.error('LSP server had an error: ' + event);
```
JonasWanke:
```suggestion
    console.error(`LSP server had an error: ${event}`);
```

new-new-packages/compiler/src/compiler/rcst_to_cst.rs
```diff
@@ -0,0 +1,229 @@
...
+                state.offset += identifier.len();
+                CstKind::Identifier(identifier)
+            }
+            Rcst::Symbol(symbol) => {
+                state.offset += symbol.len();
+                CstKind::Symbol(symbol)
+            }
+            Rcst::Int(int) => {
+                let int_str = format!("{}", int);
+                state.offset += int_str.len();
```
JonasWanke: AFAICT, we support leading zeros in integers. Using the length of the formatted value doesn't necessarily match the source length.

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+                self
+            } else {
+                Rcst::TrailingWhitespace {
+                    child: Box::new(self),
+                    whitespace,
+                }
+            }
+        } else {
+            self
+        }
```
JonasWanke: Making this a guard clause improves readability:

```suggestion
        if whitespace.is_empty() {
            return self;
        }

        if let Rcst::TrailingWhitespace {
            whitespace: self_whitespace,
            ..
        } = &mut self
        {
            self_whitespace.append(&mut whitespace);
            self
        } else {
            Rcst::TrailingWhitespace {
                child: Box::new(self),
                whitespace,
            }
        }
```

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+
+    use crate::compiler::string_to_rcst::whitespace_indentation_score;
+
+    use super::super::rcst::{IsMultiline, Rcst, RcstError};
+    use itertools::Itertools;
+
+    static MEANINGFUL_PUNCTUATION: &'static str = "=:,(){}[]->";
+
+    fn literal<'a>(input: &'a str, literal: &'static str) -> Option<&'a str> {
+        log::trace!("literal({:?}, {:?})", input, literal);
```
JonasWanke, line 73: Is this logging necessary? There's also more further down in this file

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+            }
+        }
+        None
+    }
+
+    /// "Word" refers to a number of characters that are not separated by
+    /// whitespace or significant punctuation. Identifiers, symbols, and ints
+    /// are words. Words may be invalid because they contain non-ascii or
+    /// non-alphanumeric characters – for example, the word `Magic✨` is an
+    /// invalid identifier or symbol.
```
JonasWanke:
```suggestion
    /// invalid symbol.
```

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+            Some((" blub", "I💖Candy".to_string()))
+        );
+        assert_eq!(word("012🔥hi"), Some(("", "012🔥hi".to_string())));
+        assert_eq!(word("foo(blub)"), Some(("(blub)", "foo".to_string())));
+    }
+
+    fn identifier(input: &str) -> Option<(&str, Rcst)> {
+        log::trace!("identifier({:?})", input);
+        let (input, w) = word(input)?;
+        if w.chars().next().unwrap().is_lowercase() {
```
JonasWanke: This can be simplified by converting it to a guard clause

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+                    error: RcstError::IdentifierContainsNonAlphanumericAscii,
+                }
+            ))
+        );
+    }
+
+    fn symbol(input: &str) -> Option<(&str, Rcst)> {
+        log::trace!("symbol({:?})", input);
+        let (input, w) = word(input)?;
+        if w.chars().next().unwrap().is_uppercase() {
```
JonasWanke: This can be simplified by converting it to a guard clause

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+                    error: RcstError::SymbolContainsNonAlphanumericAscii,
+                }
+            ))
+        );
+    }
+
+    fn int(input: &str) -> Option<(&str, Rcst)> {
+        log::trace!("int({:?})", input);
+        let (input, w) = word(input)?;
+        if w.chars().next().unwrap().is_ascii_digit() {
```
JonasWanke: This can be simplified by converting it to a guard clause

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+                    },
+                ))
+            }
+        } else {
+            None
+        }
+    }
+    #[test]
+    fn test_int() {
+        assert_eq!(int("42 "), Some((" ", Rcst::Int(42))));
```
JonasWanke:
```suggestion
        assert_eq!(int("42 "), Some((" ", Rcst::Int(42))));
        assert_eq!(int("012"), Some(("", Rcst::Int(12))));
```

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+        log::trace!("single_line_whitespace({:?})", input);
+        let mut chars = vec![];
+        let mut has_error = false;
+        while let Some(c) = input.chars().next() {
+            match c {
+                ' ' => {
+                    chars.push(' ');
+                    input = &input[1..];
+                }
+                c if c.is_whitespace() && c != '\n' && c != '\r' => {
```
JonasWanke: `is_whitespace` also matches, e.g., Unicode's paragraph separator (https://doc.rust-lang.org/std/primitive.char.html#method.is_whitespace). Maybe we should whitelist the whitespace we support?

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+        );
+    }
+
+    fn comment(input: &str) -> Option<(&str, Rcst)> {
+        log::trace!("comment({:?})", input);
+        let (mut input, octothorpe) = octothorpe(input)?;
+        let mut comment = vec![];
+        loop {
+            match input.chars().next() {
+                Some('\n') | None => {
```
JonasWanke: I think `\r` should end a comment as well

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+            };
+            chars.push(c);
+            has_weird_whitespace |= is_weird;
+            indent_in_spaces += indent_bonus;
+            input = &input[c.len_utf8()..];
+        }
+        let whitespace = chars.into_iter().join("");
+        Some(if has_weird_whitespace {
+            (
+                input,
```
JonasWanke: This can be lifted out of the `if`

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+            indentation,
+            also_comments
+        );
+        let mut parts = vec![];
+        if let Some((new_input, whitespace)) = single_line_whitespace(input) {
+            input = new_input;
+            parts.push(whitespace);
+        }
+
+        let mut input = input;
```
JonasWanke: Is this really necessary? You can already reassign the `input` parameter (as you already do above)

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+            while let Some((new_new_input, newline)) = newline(new_input) {
+                new_parts.push(newline);
+                new_input = new_new_input;
+            }
+            if new_input == input {
+                break; // No newlines.
+            }
+            match leading_indentation(new_input, indentation) {
+                Some((new_input, whitespace)) => {
+                    new_parts.push(Rcst::Whitespace(whitespace.to_string()));
```
JonasWanke: `whitespace` should already be an `Rcst::Whitespace`

```suggestion
                    new_parts.push(whitespace);
```

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+            let mut new_input = input;
+            let mut new_parts = vec![];
+            while let Some((new_new_input, newline)) = newline(new_input) {
+                new_parts.push(newline);
+                new_input = new_new_input;
+            }
+            if new_input == input {
+                break; // No newlines.
+            }
+            match leading_indentation(new_input, indentation) {
```
JonasWanke: We might also want to match lines that contain insufficient indentation but nothing else. For example, I can imagine the following happening often during editing, where `|` represents the cursor (indented only one space):

```candy
foo =
  call 1
 |
  call 2
```

Currently, this is a parser error (if I understand the code correctly), but that might be confusing during editing.

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+                    parts.push(Rcst::TextPart(line.drain(..).join("")));
+                    let (i, mut whitespace) =
+                        whitespaces_and_newlines(input, indentation + 1, false);
+                    input = i;
+                    parts.append(&mut whitespace);
+                    if let Some('\n') = input.chars().next() {
+                        break Rcst::Error {
+                            unparsable_input: "".to_string(),
+                            error: RcstError::TextNotSufficientlyIndented,
+                        };
```
JonasWanke, line 609: As with `whitespaces_and_newlines`, we might also want to match lines that contain insufficient indentation but nothing else

JonasWanke, line 609: As with `whitespaces_and_newlines`, we might also want to match lines that contain insufficient indentation but nothing else

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+                }
+            })
+            .or_else(|| identifier(input))
+            .or_else(|| {
+                word(input).map(|(input, word)| {
+                    (
+                        input,
+                        Rcst::Error {
+                            unparsable_input: word,
+                            error: RcstError::UnexpectedPunctuation,
```
JonasWanke: Maybe rather `UnexpectedCharacters`?

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+        let mut has_multiline_whitespace = false;
+        loop {
+            let (i, whitespace) = whitespaces_and_newlines(input, indentation + 1, true);
+            has_multiline_whitespace |= whitespace.is_multiline();
+            let indentation = if has_multiline_whitespace {
+                indentation + 1
+            } else {
+                indentation
+            };
+            log::debug!("Got whitespace: {:?}", whitespace);
```
JonasWanke: Is this still necessary?

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+        // .. # this is indentation
+        assert_eq!(
+            call("foo\n  ", 0),
+            Some((
+                "",
+                Rcst::Call {
+                    name: Box::new(Rcst::Identifier("foo".to_string())),
+                    arguments: vec![],
+                },
+            ))
```
JonasWanke: Where does the whitespace appear in the RCST?

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+    }
+
+    fn call(input: &str, indentation: usize) -> Option<(&str, Rcst)> {
+        log::trace!("call({:?}, {:?})", input, indentation);
+        let (input, mut expressions) = run_of_expressions(input, indentation)?;
+        if expressions.len() < 2 {
+            return None;
+        }
+        let arguments = expressions.split_off(1);
+        let name = expressions.into_iter().next().unwrap();
```
JonasWanke, line 816: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.first

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+        assert_eq!(
+            call("foo 1 2\n  3\n  4\nbar", 0),
+            Some((
+                "\nbar",
+                Rcst::Call {
+                    name: Box::new(Rcst::Identifier("foo".to_string())),
+                    arguments: vec![Rcst::Int(1), Rcst::Int(2), Rcst::Int(3), Rcst::Int(4)],
+                }
+            ))
+        );
```
JonasWanke: What about the whitespace?

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+        let (input, mut opening_curly_brace) = opening_curly_brace(input)?;
+        let (mut input, mut parameters_and_arrow) = {
+            let input_without_params = input;
+            let mut input = input;
+            let mut parameters = vec![];
+            loop {
+                let (i, whitespace) = whitespaces_and_newlines(input, indentation + 1, true);
+                if parameters.is_empty() {
+                    opening_curly_brace = opening_curly_brace.wrap_in_whitespace(whitespace);
+                }
```
JonasWanke: If there are already some parameters, the whitespace is silently discarded

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+                    opening_curly_brace: Box::new(Rcst::OpeningCurlyBrace),
+                    parameters_and_arrow: Some((
+                        vec![Rcst::Identifier("a".to_string())],
+                        Box::new(Rcst::Arrow)
+                    )),
+                    body: vec![Rcst::Identifier("foo".to_string())],
+                    closing_curly_brace: Box::new(Rcst::ClosingCurlyBrace),
+                }
+            ))
+        );
```
JonasWanke: Whitespace is missing in the RCST

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+        if signature.is_empty() {
+            return None;
+        }
+
+        let (input, whitespace) = whitespaces_and_newlines(input, indentation + 1, true);
+        let last = signature.pop().unwrap();
+        signature.push(last.wrap_in_whitespace(whitespace.clone()));
+
+        let parameters = signature.split_off(1);
+        let name = signature.into_iter().next().unwrap();
```
JonasWanke, line 1523: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.first

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+            Some((
+                "",
+                Rcst::Assignment {
+                    name: Box::new(Rcst::Identifier("foo".to_string())),
+                    parameters: vec![],
+                    equals_sign: Box::new(Rcst::EqualsSign),
+                    body: vec![Rcst::Int(42)],
+                }
+            ))
+        );
```
JonasWanke: Whitespace is missing in the RCST

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -0,0 +1,1394 @@
...
+            Some((
+                "",
+                Rcst::Assignment {
+                    name: Box::new(Rcst::Identifier("foo".to_string())),
+                    parameters: vec![Rcst::Identifier("bar".to_string())],
+                    equals_sign: Box::new(Rcst::EqualsSign),
+                    body: vec![Rcst::Int(3)],
+                }
+            ))
+        );
```
JonasWanke: Whitespace is missing in the RCST

JonasWanke (changes requested):
Review is WIP

new-new-packages/compiler/src/compiler/ast_to_hir.rs
```diff
@@ -21,10 +21,7 @@ pub trait AstToHir: CstDb + CstToAst {
     fn cst_to_hir_id(&self, input: Input, id: cst::Id) -> Option<hir::Id>;
 
     fn hir(&self, input: Input) -> Option<(Arc<Body>, HashMap<hir::Id, ast::Id>)>;
-    fn hir_raw(
-        &self,
-        input: Input,
-    ) -> Option<(Arc<Body>, HashMap<hir::Id, ast::Id>, Vec<CompilerError>)>;
+    fn hir_raw(&self, input: Input) -> Option<(Arc<Body>, HashMap<hir::Id, ast::Id>)>;
```
JonasWanke: `hir` and `hir_raw` now return the same content, so we can get rid of `hir_raw`

new-new-packages/compiler/src/compiler/ast_to_hir.rs
```diff
@@ -383,12 +393,20 @@ impl<'c> Compiler<'c> {
...
+    fn ast_id_to_span(&self, id: &ast::Id) -> Range<usize> {
+        self.context
+            .db
+            .find_cst(
+                id.input.clone(),
+                self.context.db.ast_to_cst_id(id.clone()).unwrap(),
+            )
+            .span
+            .clone()
+    }
```
JonasWanke: `ast_id_to_span` already exists in `cst_to_ast.rs` as a query

new-new-packages/compiler/src/compiler/error.rs
```diff
@@ -1,7 +1,16 @@
 use std::ops::Range;
 
+use super::{ast::AstError, hir::HirError, rcst::RcstError};
+
 #[derive(Debug, PartialEq, Eq, Clone, Hash)]
 pub struct CompilerError {
     pub span: Range<usize>,
```
JonasWanke, line 9: In addition to the span, we also need the Input that it belongs to

new-new-packages/compiler/src/compiler/hir.rs
```diff
@@ -120,8 +125,14 @@ pub enum Expression {
...
     Body(Body),
-    Call { function: Id, arguments: Vec<Id> },
-    Error,
+    Call {
+        function: Id,
+        arguments: Vec<Id>,
+    },
+    Error {
+        child: Option<Id>,
+        errors: Vec<CompilerError>,
```
JonasWanke, line 134: I think it always contains just a single error, so the vec is unnecessary

JonasWanke, line 134: I think it always contains just a single error, so the vec is unnecessary

new-new-packages/compiler/src/compiler/cst_to_ast.rs
```diff
@@ -2,26 +2,27 @@ use std::ops::Range;
...
     fn ast_id_to_span(&self, id: ast::Id) -> Option<Range<usize>>;
 
     fn cst_to_ast_id(&self, input: Input, id: cst::Id) -> Option<ast::Id>;
 
     fn ast(&self, input: Input) -> Option<(Arc<Vec<Ast>>, HashMap<ast::Id, cst::Id>)>;
-    fn ast_raw(
-        &self,
-        input: Input,
-    ) -> Option<(Arc<Vec<Ast>>, HashMap<ast::Id, cst::Id>, Vec<CompilerError>)>;
+    fn ast_raw(&self, input: Input) -> Option<(Arc<Vec<Ast>>, HashMap<ast::Id, cst::Id>)>;
```
JonasWanke: These are now identical and we can remove `ast_raw`

new-new-packages/compiler/src/compiler/cst_to_ast.rs
```diff
@@ -42,209 +43,304 @@ fn cst_to_ast_id(db: &dyn CstToAst, input: Input, id: cst::Id) -> Option<ast::Id
...
+                            ..
+                        } => Some(text),
+                        _ => panic!("Text contains non-TextPart. Whitespaces should have been removed already."),
+                    })
+                    .join("");
+                let string = self.create_string_without_id_mapping(text);
+                let mut text = self.create_ast(cst.id, AstKind::Text(Text(string)));
+
+                if !matches!(closing_quote.kind, CstKind::DoubleQuote) {
+                    text = self.create_ast(
```
JonasWanke, line 132: Is it intentional that the text itself is now discarded?

new-new-packages/compiler/src/compiler/cst_to_ast.rs
```diff
@@ -42,209 +43,304 @@ fn cst_to_ast_id(db: &dyn CstToAst, input: Input, id: cst::Id) -> Option<ast::Id
...
-                "Expected a closing parenthesis to end a parenthesized expression, but found `{}`.",
-                *closing_parenthesis
-            );
-                self.lower_cst(inner)
+                    matches!(opening_parenthesis.kind, CstKind::OpeningParenthesis),
+                    "Parenthesized needs to start with opening parenthesis, but started with {}.",
+                    opening_parenthesis
+                );
+                if !matches!(closing_parenthesis.kind, CstKind::ClosingParenthesis) {
+                    ast = self.create_ast(
```
JonasWanke, line 163: Is it intentional that the inner is now discarded?

new-new-packages/compiler/src/compiler/cst_to_ast.rs
```diff
@@ -42,209 +43,304 @@ fn cst_to_ast_id(db: &dyn CstToAst, input: Input, id: cst::Id) -> Option<ast::Id
...
+                            }],
+                        },
+                    );
+                }
+
+                ast
+            }
+            CstKind::Call { name, arguments } => {
+                let name_string = if let CstKind::Identifier(identifier) = &name.kind {
+                    Some(self.create_string(cst.id.to_owned(), identifier.to_owned()))
```
JonasWanke: I think this should use the `name`'s CST ID, though I'm not sure

new-new-packages/compiler/src/compiler/cst_to_ast.rs
```diff
@@ -42,209 +43,304 @@ fn cst_to_ast_id(db: &dyn CstToAst, input: Input, id: cst::Id) -> Option<ast::Id
...
+                                    AstKind::Error {
+                                        child: None,
+                                        errors: vec![CompilerError {
+                                            span: colon.span.clone(),
+                                            payload: CompilerErrorPayload::Ast(
+                                                AstError::ColonMissingAfterStructKey,
+                                            ),
+                                        }],
+                                    },
+                                )
```
JonasWanke: Is it intentional that the key itself is now discarded?

new-new-packages/compiler/src/compiler/cst_to_ast.rs
```diff
@@ -42,209 +43,304 @@ fn cst_to_ast_id(db: &dyn CstToAst, input: Input, id: cst::Id) -> Option<ast::Id
...
+                                            child: None,
+                                            errors: vec![CompilerError {
+                                                span: comma.span.clone(),
+                                                payload: CompilerErrorPayload::Ast(
+                                                    AstError::NonCommaAfterStructValue,
+                                                ),
+                                            }],
+                                        },
+                                    )
+                                }
```
JonasWanke: Is it intentional that the value itself is now discarded?

new-new-packages/compiler/src/compiler/cst_to_ast.rs
```diff
@@ -270,49 +364,64 @@ impl LoweringContext {
...
+            .into_iter()
             .enumerate()
-            .map(|(index, it)| {
-                self.lower_parameter(it)
-                    .unwrap_or_else(|| self.create_string(it.id, format!("<invalid#{}", index)))
+            .map(|(index, it)| match self.lower_parameter(it) {
+                Ok(parameter) => parameter,
+                Err(error) => {
+                    errors.push(error);
+                    self.create_string(it.id, format!("<invalid#{}", index))
```
JonasWanke:
```suggestion
                    self.create_string(it.id, format!("<invalid#{}>", index))
```

new-new-packages/compiler/src/language_server/folding_range.rs
```diff
@@ -42,34 +43,50 @@ impl<'a> Context<'a> {
...
+            CstKind::Newline(_) => {}
             // TODO: support folding ranges for comments
-            CstKind::LeadingComment { child, .. } => self.visit_cst(child),
+            CstKind::Comment { .. } => {}
             CstKind::TrailingWhitespace { child, .. } => self.visit_cst(child),
-            CstKind::TrailingComment { child, .. } => self.visit_cst(child),
+            CstKind::Identifier(_) => {}
+            CstKind::Symbol(_) => {}
+            CstKind::Int(_) => {}
+            CstKind::Text { .. } => {}
```
JonasWanke, line 67: We could add a TODO for multi-line texts

new-new-packages/compiler/src/language_server/folding_range.rs
```diff
@@ -42,34 +43,50 @@ impl<'a> Context<'a> {
...
+                        name.span.end,
+                        last_argument.span.end,
+                        FoldingRangeKind::Region,
+                    );
                 }
+
+                self.visit_cst(name);
+                self.visit_csts(&arguments);
+            }
+            CstKind::Struct { fields, .. } => self.visit_csts(fields),
```
JonasWanke, line 88: We could add a TODO for structs

new-new-packages/compiler/src/language_server/semantic_tokens.rs
```diff
@@ -153,42 +154,35 @@ impl<'a> Context<'a> {
...
-                self.add_token(span.end - value.len()..span.end, SemanticTokenType::Comment);
+                self.visit_cst(child, token_type_for_identifier)
             }
+            CstKind::Identifier { .. } => self.add_token(
+                cst.span.clone(),
+                token_type_for_identifier.unwrap_or(SemanticTokenType::Function),
+            ),
+            CstKind::Symbol { .. } => self.add_token(cst.span.clone(), SemanticTokenType::Symbol),
+            CstKind::Int { .. } => self.add_token(cst.span.clone(), SemanticTokenType::Number),
+            CstKind::Text { .. } => self.add_token(cst.span.clone(), SemanticTokenType::String),
```
JonasWanke: Now that linebreaks and indentation are not part of the text, maybe we should only report the actual `TextPart`s and quotes?

new-new-packages/compiler/src/main.rs
```diff
@@ -52,83 +58,131 @@ struct CandyRunOptions {
...
+        let mut watcher = watcher(tx, Duration::from_secs(1)).unwrap();
+        watcher
+            .watch(&options.file, RecursiveMode::Recursive)
+            .unwrap();
+        loop {
+            match rx.recv() {
+                Ok(_) => {
+                    raw_build(&options.file, options.debug);
+                }
+                Err(e) => println!("watch error: {:#?}", e),
```
JonasWanke:
```suggestion
                Err(e) => log::error!("watch error: {:#?}", e),
```

JonasWanke (changes requested):

new-new-packages/compiler/src/compiler/ast.rs
MarcelGarus: If the value is multiple lines, all lines should be indented.

For example, we want this:

```
[
  Foo: Bar,
  Baz: [
    Foo: Bar,
  ],
]
```

Instead of this:

```
[
  Foo: Bar,
  Baz: [
  Foo: Bar
],
]
```

new-new-packages/compiler/src/compiler/ast.rs
MarcelGarus: Same here

new-new-packages/compiler/src/compiler/string_to_rcst.rs
MarcelGarus: We need the `Rcst`, but there's only `.first()` returning `&Rcst` and `.first_mut()` returning `&mut Rcst`. There would also be a nice `.split_first()` method on `[T]` to do `let (name, parameters) = signature.split_first();`, but it has the same problem.

new-new-packages/compiler/src/compiler/hir.rs
MarcelGarus: It can contain multiple errors, in particular multiple errors from lower levels.
For example, if a non-identifier is called, an `AstError::CallOfANonIdentifier` is generated, but all errors of the arguments are collected as well and returned in the same error node.

new-new-packages/compiler/src/compiler/ast.rs
MarcelGarus: And here

new-new-packages/compiler/src/compiler/hir.rs
MarcelGarus: It's quite the opposite: Before there were only commas between items, now there's a comma after every item.

new-new-packages/compiler/src/compiler/cst_to_ast.rs
```diff
@@ -452,7 +459,7 @@ impl LoweringContext {
         id
     }
     fn create_next_id_without_mapping(&mut self) -> ast::Id {
-        let id = ast::Id::new(self.input.clone(), self.next_id);
+        let id = ast::Id::new(self.input.clone().clone(), self.next_id);
```
JonasWanke: Is it really necessary to clone the clone?

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -417,35 +442,45 @@ mod parse {
...
-                    parts.append(&mut new_parts);
-                    input = new_input;
-                }
-                None => break,
+
+            if new_input == new_input_from_iteration_start {
+                break;
             }
         }
+
```
JonasWanke, line 484: Could you add a test for indentation 1 with `# foo\n\n  # bar\n`?

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -1188,32 +1305,53 @@ mod parse {
...
+            if let Some((i, curly_brace)) = closing_curly_brace(i) {
+                (i, body_expression, whitespace, curly_brace)
+            } else {
+                // There is not a closing brace after a single expression. Try a
+                // body of multiple expressions. We didn't try this first
+                // because then the body would also have consumed any trailing
+                // closing curly brace in the same line.
+                // For example, for the lambda `{ 2 }`, the body parser would
+                // have already consumed the `}`. The body parser works great
+                // for multiline bodies though.
```
JonasWanke:
```suggestion
                // for multiline bodies, though.
```

new-new-packages/compiler/src/compiler/string_to_rcst.rs
```diff
@@ -1233,9 +1371,18 @@ mod parse {
             Some((
```
JonasWanke, line 1394: Can you please add a test for `{ foo\n  bar\n}`?

JonasWanke (changes requested):

JonasWanke (approved):

JonasWanke (approved):

PR #228, opened 2022-07-07
Add support for use and useAsset

MarcelGarus:


MarcelGarus:
Oh, import cycles are not detected yet.

compiler/src/compiler/cst.rs
```diff
@@ -222,7 +229,7 @@ impl Display for Cst {
             CstKind::Assignment {
                 name,
                 parameters,
-                equals_sign,
+                assignment_sign: equals_sign,
```
JonasWanke: Nit: Can you please update the code below to also use `assignment_sign`?

compiler/src/compiler/cst.rs
```diff
@@ -342,12 +349,12 @@ impl UnwrapWhitespaceAndComment for Cst {
...
```
JonasWanke: Nit: Can you please update the code below to also use `assignment_sign`?

compiler/src/compiler/cst.rs
```diff
@@ -466,7 +474,7 @@ impl TreeWithIds for Cst {
...
```
JonasWanke: Nit: Can you please update the code below to also use `assignment_sign`?

compiler/src/compiler/cst.rs
```diff
@@ -544,7 +553,7 @@ impl TreeWithIds for Cst {
...
```
JonasWanke: Nit: Can you please update the code below to also use `assignment_sign`?

compiler/src/compiler/rcst.rs
```diff
@@ -215,7 +217,7 @@ impl Display for Rcst {
             Rcst::Assignment {
                 name,
                 parameters,
-                equals_sign,
+                assignment_sign: equals_sign,
```
JonasWanke: Nit: Can you please update the code below to also use `assignment_sign`?

compiler/src/compiler/rcst.rs
```diff
@@ -327,7 +330,7 @@ impl IsMultiline for Rcst {
...
```
JonasWanke: Nit: Can you please update the code below to also use `assignment_sign`?

compiler/src/input.rs
```diff
@@ -2,35 +2,44 @@ use crate::database::PROJECT_DIRECTORY;
...
 
     match input {
         Input::File(_) | Input::ExternalFile(_) => {
             let path = input.to_path().unwrap();
-            match read_to_string(path.clone()) {
+            match fs::read(path.clone()) {
                 Ok(content) => Some(Arc::new(content)),
                 Err(error) if matches!(error.kind(), std::io::ErrorKind::NotFound) => None,
-                _ => panic!("Unexpected error when reading file {:?}.", path),
+                // _ => panic!("Unexpected error when reading file {:?}.", path),
```
JonasWanke: Can we at least log this?

compiler/src/compiler/string_to_rcst.rs
```diff
@@ -1613,11 +1629,14 @@ mod parse {
         let parameters = signature.split_off(1);
         let name = signature.into_iter().next().unwrap();
 
-        let (input, mut equals_sign) = equals_sign(input)?;
-        let input_after_equals_sign = input;
+        let (input, mut assignment_sign) = match colon_equals_sign(input) {
+            Some(it) => it,
+            None => equals_sign(input)?,
+        };
```
JonasWanke: Can we use `option.or_else(fn_once)` here or does that cause ownership problems?

packages/Core/Panic.candy
```diff
@@ -1 +1 @@
-panic message = ✨.panic message
+panic := { message -> ✨.panic message }
```
JonasWanke: Why do you define this as a lambda instead of a function?

compiler/src/vm/builtin_functions.rs
```diff
@@ -165,148 +183,139 @@ impl Vm {
...
+            content
+                .iter()
+                .map(|byte| Value::Int(*byte as u64))
+                .collect_vec(),
+        ))
+    }
+
+    fn use_local_module(&mut self, db: &Database, args: Vec<Value>) -> Result<(), String> {
+        let (current_path, target) = Self::parse_current_path_and_target(args)?;
+        let target = UseAssetTarget::parse(&target)?;
```
JonasWanke: If it's also used for local modules, maybe `UseAssetTarget` should rather be called `UseTarget`?

compiler/src/vm/builtin_functions.rs
```diff
@@ -165,148 +183,139 @@ impl Vm {
...
+        let possible_inputs = target.resolve_local_module(&current_path)?;
+        let input = 'find_existing_input: {
+            for input in possible_inputs {
+                if db.get_input(input.clone()).is_some() {
+                    break 'find_existing_input input;
+                }
+            }
+            return Err("couldn't import module".to_string());
+        };
```
JonasWanke: Isn't something like the following possible?

```suggestion
        let input = possible_inputs.iter()
            .filter_map(|it| db.get_input(it.clone()))
            .next()
            .or_or_else("couldn't import module".to_string())?;
```

compiler/src/vm/builtin_functions.rs
```diff
@@ -165,148 +183,139 @@ impl Vm {
...
+        if path
+            .last()
+            .map(|it| it.ends_with(".candy"))
+            .unwrap_or(false)
+        {
+            return Err(
+                "importing child files (starting with a single dot) only works from `.candy` files"
+                    .to_string(),
+            );
         }
```
JonasWanke: I think the condition and error message don't quite match. Maybe the condition should use `current_path` and also check that `self.parent_navigations == 0`?

compiler/src/builtin_functions.rs
```diff
@@ -4,17 +4,18 @@ use strum_macros::EnumIter;
...
+    GetArgumentCount, // closure -> argumentCount
+    IfElse,           // condition thenClosure elseClosure -> resultOfExecutedClosure
+    Panic,            // message -> Never
+    Print,            // message -> Nothing
+    StructGet,        // struct key -> value
+    StructGetKeys,    // struct -> listOfKeys
+    StructHasKey,     // struct key -> bool
+    TypeOf,           // any -> typeSymbol
+    UseAsset,         // currentPath target -> targetAsString
+    UseLocalModule,   // currentPath target -> targetAsStruct
```
JonasWanke:
```suggestion
    Add,              // (summandA: int) (summandB: int) -> (sum: int)
    Equals,           // (a: any) (b: any) -> (boolean: True | False)
    GetArgumentCount, // (closure: Lambda) -> (argumentCount: int)
    IfElse,           // (condition: True | False) (then: lambda) (else: lambda) -> (resultOfExecutedClosure: any)
    Panic,            // (message: Text) -> Never
    Print,            // (message: Text) -> Nothing
    StructGet,        // (struct: struct) (key: any) -> (value: any)
    StructGetKeys,    // (struct: struct) -> (listOfKeys: listOfAny)
    StructHasKey,     // (struct: struct) (key: any) -> (isKeyInStruct: True | False)
    TypeOf,           // (value: any) -> (type: Int | Text | Symbol | Struct | Function | Builtin)
    UseAsset,         // (currentPath: struct) (target: text) -> (target: text)
    UseLocalModule,   // (currentPath: struct) (target: text) -> (target: struct)
```

compiler/src/vm/heap.rs
```diff
@@ -1,39 +1,59 @@
...
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct Heap {
     objects: HashMap<ObjectPointer, Object>,
     next_address: ObjectPointer,
 }
 
 pub type ObjectPointer = usize;
 
-#[derive(Debug, Clone)] // TODO: remove Clone once it's no longer needed
```
JonasWanke, line 14: Why are these TODOs gone (this and the one below on line 19)?

compiler/src/vm/value.rs
```diff
@@ -49,19 +75,19 @@ impl Value {
...
                 entries
                     .iter()
                     .map(|(key, value)| format!("{}: {}", key, value))
                     .join(", ")
             ),
-            Value::Closure { body, .. } => {
-                write!(f, "{{{body}}}")
+            Value::Closure { .. } => {
+                write!(f, "{{...}}")
             }
```
JonasWanke:
```suggestion
            Value::Closure { .. } => write!(f, "{{...}}"),
```

compiler/src/vm/vm.rs
```diff
@@ -14,46 +15,115 @@ use std::collections::HashMap;
...
+
+        for arg in arguments {
+            let address = self.heap.import(arg);
+            self.data_stack.push(address);
+        }
+        let address = self.heap.import(closure);
+        self.data_stack.push(address);
+
+        self.run_instruction(db, Instruction::Call { num_args });
+        self.status = Status::Running;
```
JonasWanke, line 97: Shouldn't we set the status to running before executing the instruction?

compiler/src/vm/vm.rs
```diff
@@ -14,46 +15,115 @@ use std::collections::HashMap;
...
+        } = closure.clone()
+        {
+            assert_eq!(captured.len(), 0, "Called start_module_closure with a closure that is not a module closure (it captures stuff).");
+            assert_eq!(num_args, 0, "Called start_module_closure with a closure that is not a module closure (it has arguments).");
+            body
+        } else {
+            panic!("Called start_module_closure with a non-closure.");
+        };
+        instructions.insert(0, Instruction::TraceModuleStarts { input });
+        instructions.push(Instruction::TraceModuleEnds);
```
JonasWanke: `Value::module_closure_of_input(…)` already inserts these trace markers

compiler/src/vm/vm.rs
```diff
@@ -171,14 +268,12 @@ impl Vm {
                             self.heap.dup(captured);
                         }
                         self.data_stack.append(&mut args);
-                        self.heap.drop(closure_address);
```
JonasWanke, line 174: Was this line removed intentionally?

compiler/src/vm/vm.rs
```diff
@@ -252,10 +348,33 @@ impl Vm {
...
+            Instruction::TraceModuleStarts { input } => {
+                if self.import_stack.contains(&input) {
+                    self.panic(format!(
+                        "there's an import cycle ({})",
+                        self.import_stack
+                            .iter()
+                            .skip_while(|it| **it != input)
+                            .chain([&input])
+                            .map(|input| format!("{input}"))
+                            .join(" -> "),
```
JonasWanke:
```suggestion
                            .join(" → "),
```

compiler/src/fuzzer/mod.rs
```diff
@@ -111,12 +110,14 @@ fn test_closure_with_args(
     closure_address: usize,
     arguments: Vec<Value>,
 ) -> TestResult {
-    vm.run_closure(closure_address, arguments);
+    let closure = vm.heap.export_without_dropping(closure_address);
+    println!("Starting closure {closure}.");
```
JonasWanke:
```suggestion
    log::trace!("Starting closure {closure}.");
```

compiler/src/compiler/hir.rs
```diff
@@ -116,6 +117,10 @@ impl Id {
             }),
         }
     }
+
+    pub fn is_parent_of(&self, other: &Self) -> bool {
```
JonasWanke: This detects direct and indirect parents. If this is intentional, can we make this clearer in the name, maybe `is_any_parent_of`?

compiler/src/compiler/ast_to_hir.rs
```diff
@@ -314,22 +212,50 @@ impl<'c> Compiler<'c> {
...
                     Some(name.id.clone()),
-                    Expression::Reference(body),
-                    Some(name_string),
-                )
+                    Expression::Reference(body.clone()),
+                    Some(name_string.clone()),
+                );
+                if *is_public {
+                    if self.is_top_level {
+                        self.public_identifiers.insert(name_string, body.clone());
```
JonasWanke, line 242: We should probably check whether a public identifier with the same name already exists and add an error in that case

compiler/src/compiler/ast_to_hir.rs
```diff
@@ -510,6 +434,7 @@ impl<'c> Compiler<'c> {
         expression: Expression,
         identifier: Option<String>,
     ) -> hir::Id {
+        let identifier: Option<_> = identifier.into();
```
JonasWanke: What is this line doing? `identifier` already is an `Option` before this line, so presumably the type argument changes. However, you appear to then be using it as an `Option<String>`, so I think it's not actually changing the type

```suggestion
```

compiler/src/compiler/ast_to_hir.rs
```diff
@@ -542,6 +464,215 @@ impl<'c> Compiler<'c> {
...
+                        function: panic_id,
+                        arguments: vec![message_id],
+                    },
+                    Some("panicked".to_string()),
+                )
+            }
+            Input::Untitled(_) => {
+                let message_id = self.push(
+                    None,
+                    Expression::Text("Untitled files can't call `useAsset`.".to_string()),
```
JonasWanke:
```suggestion
                    Expression::Text("Untitled files can't call `use` or `useAsset`.".to_string()),
```

compiler/src/compiler/ast_to_hir.rs
```diff
@@ -542,6 +464,215 @@ impl<'c> Compiler<'c> {
...
+                    Some("panicked".to_string()),
+                )
+            }
+        }
+    }
+
+    fn generate_use_asset(&mut self) {
+        // HirId(~:test.candy:useAsset) = lambda { HirId(~:test.candy:useAsset:target) ->
+        //   HirId(~:test.candy:useAsset:panic) = builtinPanic
+        //   HirId(~:test.candy:useAsset:useAsset) = builtinUseAsset
```
JonasWanke: This is generated after `currentPath`

compiler/src/compiler/ast_to_hir.rs
```diff
@@ -542,6 +464,215 @@ impl<'c> Compiler<'c> {
...
+        //     HirId(~:test.candy:useAsset:target)
+        // }
+
+        let reset_state = self.start_scope();
+        self.prefix_keys.push("useAsset".to_string());
+
+        let lambda_parameter_id = hir::Id::new(
+            self.input.clone(),
+            add_keys(&self.prefix_keys[..], "target".to_string()),
+        );
```
JonasWanke: Move the empty line to unify this and `generate_use(…)`

```suggestion
        let lambda_parameter_id = hir::Id::new(
            self.input.clone(),
            add_keys(&self.prefix_keys[..], "target".to_string()),
        );

```

JonasWanke (changes requested):

packages/Core/Panic.candy
MarcelGarus: Good question. I thought this would have made some sense, but I can't remember why.
Another idea: What do you think about implementing panic like this?

```candy
panic message :=
  needs (isText message) "panic needs a text as the message"
  needs False message
```

compiler/src/vm/builtin_functions.rs
MarcelGarus: True, not sure what happened here.

compiler/src/builtin_functions.rs
MarcelGarus: Looks good! Although the Rust formatter messes things up

compiler/src/vm/heap.rs
MarcelGarus: I thought that it's actually pretty useful to be able to clone VMs and Objects (such as when snapshotting the VM state and fuzzing the copy without affecting the original).

compiler/src/vm/vm.rs
MarcelGarus: Yup. Right here, we are about to enter the closure body and that's not a good time to drop the closure and everything it captured.

Instead, the `Return` instruction now pops the closure from the call stack and then drops it.

packages/Core/Panic.candy
MarcelGarus: Thought about this a bit more and this would match the semantics: There's no correct way to call `panic`. You should always be warned if there's a path in your code that reaches a panic call.

MarcelGarus: Okay, I now implemented `panic` using the needs.

JonasWanke (approved):

PR #229, opened 2022-07-11
Value hints

MarcelGarus:
Depends on #228

### DIFF
--- a/compiler/src/language_server/mod.rs
+++ b/compiler/src/language_server/mod.rs
@@ -17,7 +17,7 @@ use crate::{
     compiler::{ast_to_hir::AstToHir, hir::CollectErrors},
     database::PROJECT_DIRECTORY,
     input::{Input, InputDb},
-    Database,
+    CloneWithExtension, Database,
 };
 use itertools::Itertools;
 use lsp_types::{
@@ -31,7 +31,7 @@ use lsp_types::{
     TextDocumentChangeRegistrationOptions, TextDocumentContentChangeEvent,
     TextDocumentRegistrationOptions, Url, WorkDoneProgressOptions,
 };
-use std::sync::Arc;
+use std::{fs, sync::Arc};
 use tokio::sync::{mpsc::Sender, Mutex};
 use tower_lsp::{jsonrpc, Client, LanguageServer};
 
@@ -46,6 +46,14 @@ impl CandyLanguageServer {
             client,
             db: Default::default(),
             hints_server_sink: Default::default(),
+        }
+    }
+
+    async fn send_to_hints_server(&self, event: hints::Event) {
+        let event_sink = self.hints_server_sink.lock().await;
+        match event_sink.as_ref().unwrap().send(event).await {
+            Ok(_) => {}
+            Err(_) => panic!("Couldn't shutdown hints server."),
         }
     }
 }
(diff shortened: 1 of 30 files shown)


compiler/src/language_server/mod.rs
```diff
@@ -25,25 +31,29 @@ use lsp_types::{
...
+            db: Default::default(),
+            hints_server_sink: Default::default(),
+        }
+    }
+
+    async fn send_to_hints_server(&self, event: hints::Event) {
+        let event_sink = self.hints_server_sink.lock().await;
+        match event_sink.as_ref().unwrap().send(event).await {
+            Ok(_) => {}
+            Err(_) => panic!("Couldn't shutdown hints server."),
```
JonasWanke:
```suggestion
            Err(_) => panic!("Couldn't send message to hints server."),
```

packages/Test.candy.hints
```diff
@@ -0,0 +1,5 @@
+Hint { kind: Value, text: " # [ 0: 110, 10: 97, 16: 50, 20: 111, 24: 44, 4: 44, 6: 103, 3: 101, 7: 101, 13: 101, 22: 97, 21: 110, 9: 77, 23: 115, 17: 50, 14: 108, 19: 74, 5: 97, 8: 10, 11: 114, 25: 50, 12: 99, 26: 50, 18: 10, 1: 97, 15: 44, 2: 109 ]", position: Position { line: 0, character: 28 } }
```
JonasWanke, line 1: I think we can `.gitignore` the generated `*.candy.*` files

compiler/src/vm/use_provider.rs
```diff
@@ -0,0 +1,48 @@
...
+            .map(|bytes| (*bytes).clone())
+            .ok_or_else(|| format!("Couldn't import file '{}'.", input))
+    }
+
+    fn use_local_module(&self, input: Input) -> Option<Lir> {
+        self.db.lir(input).map(|lir| (*lir).clone())
+    }
+}
+
+pub struct FunctionUseProvider<'a, U1, U2>
```
JonasWanke: This does not appear to be used anywhere. Do we need it or was it just temporary?

JonasWanke (changes requested):

packages/Test.candy.hints
MarcelGarus: Oh, definitely, good idea. I have several of these files which I always consciously ignore in GitHub Desktop, but seems like one slipped through.

compiler/src/vm/builtin_functions.rs
```diff
@@ -135,7 +135,7 @@ impl Vm {
 
     fn print(&mut self, args: Vec<Value>) -> Result<Value, String> {
         destructure!(args, [Value::Text(message)], {
-            println!("{message:?}");
+            log::info!("{message:?}");
             Ok(Value::nothing())
         })
```
JonasWanke, line 140: I think this broke, maybe during a merge?

JonasWanke (changes requested):

compiler/src/vm/builtin_functions.rs
MarcelGarus: What's the problem here? This seems fine to me

JonasWanke: Interesting, this was just showing me a different diff… Then it LGTM

JonasWanke (approved):

PR #230, opened 2022-07-14
Fuzzing

MarcelGarus:
Depends on #229

This PR only implements a very simple fuzzing with a limited number of instructions. Long term, it makes sense to enable fuzzing complex functions by adding a cache and storing in-progress VMs of currently fuzzed functions. I added todos for that.


MarcelGarus:
This can already be reviewed, but before merging, I'll still try out aligning the hints.

MarcelGarus:
Okay, now this is ready for review for real. The fuzzer now works incrementally, allowing us to fuzz even functions that take a long time. And the hints are aligned (although one could argue we should display them even more on the right so that the entire function code is left of them and that there should be hints for intermediate results).

hints

README.md
```diff
@@ -95,10 +95,10 @@ Major milestones:
   * [ ] moniker
 * [x] incremental compilation
 * [ ] lists
-* [ ] maps
+* [x] maps
```
JonasWanke:
```suggestion
* [x] structs
```

compiler/src/compiler/string_to_rcst.rs
```diff
@@ -1040,6 +1041,28 @@ mod parse {
                 }
             ))
         );
+        // foo T
+        //
+        // bar = 5
```
JonasWanke: In the string, there are three newline characters

```suggestion
        // foo T
        //
        //
        // bar = 5
```

compiler/src/fuzzer/fuzzer.rs
```diff
@@ -0,0 +1,112 @@
...
+    // machine and during transitioning to a new state (aka running the fuzzer),
+    // we'd like to consume the old status and then produce a new status.
+    // Rust's ownership rules don't let us take ownership of the status (leaving
+    // it uninitialized), even if it's "just temporarily" while we're
+    // transitioning. The reason is that our state machine code could panic and
+    // in that case, some status needs to be there to be freed.
+    // In the future, we could use `unsafe` to set the status to `uninit()`. But
+    // currently, I'm not 100% that our VM won't panic, so we instead set it to
+    // this temporary value.
+    TemporarilyUninitialized,
```
JonasWanke: Maybe we should just call this `Running` or `ActivelyRunning`? Alternatively, could the `Fuzzer` store a private `status: Option<Status>` and expose this via a function that unwraps it?

compiler/src/fuzzer/mod.rs
```diff
@@ -1,194 +1,93 @@
...
-            if is_our_fault {
-                TestResult::WrongInputs
-            } else {
-                TestResult::InternalPanic(message)
-            }
+    for (id, closure) in fuzzable_closures {
+        let mut fuzzer = Fuzzer::new(db, closure.clone(), id.clone());
+        for _ in 0..20 {
+            fuzzer.run(db, 100);
         }
```
JonasWanke: Shouldn't this accomplish the same?

```suggestion
        fuzzer.run(db, 2000);
```

compiler/src/language_server/hints/constant_evaluator.rs
```diff
@@ -0,0 +1,185 @@
...
+    pub fn run(&mut self, db: &Database) -> Option<Input> {
+        let num_vms = self.vms.len();
+        let mut running_vms = self
+            .vms
+            .iter_mut()
+            .filter(|(_, vm)| match vm.status() {
+                Status::Running => true,
+                Status::Done => false,
+                Status::Panicked(_) => false,
+            })
```
JonasWanke:
```suggestion
            .filter(|(_, vm)| matches!(vm.status(), Status::Running))
```

compiler/src/language_server/hints/constant_evaluator.rs
```diff
@@ -0,0 +1,185 @@
...
+        );
+
+        running_vms.shuffle(&mut thread_rng());
+        if let Some((input, vm)) = running_vms.pop() {
+            let use_provider = DbUseProvider { db };
+            vm.run(&use_provider, 500);
+            Some(input.clone())
+        } else {
+            None
+        }
```
JonasWanke: Instead of shuffling the whole Vec and then only using the first item, can we just access a random index?

compiler/src/language_server/hints/constant_evaluator.rs
```diff
@@ -0,0 +1,185 @@
...
+        log::debug!("Calculating hints for {input}");
+        let mut hints = vec![];
+
+        match vm.status() {
+            Status::Running => {}
+            Status::Done => {}
+            Status::Panicked(value) => match panic_hint(&db, input.clone(), &vm, value) {
+                Some(hint) => {
+                    hints.push(hint);
+                }
```
JonasWanke:
```suggestion
                Some(hint) => hints.push(hint),
```

compiler/src/language_server/hints/constant_evaluator.rs
```diff
@@ -0,0 +1,185 @@
...
+                None => log::error!("Module panicked, but we are not displaying an error."),
```
JonasWanke: Maybe this could even call `.expect(…)` instead of logging an error?

compiler/src/language_server/hints/constant_evaluator.rs
```diff
@@ -0,0 +1,185 @@
...
+                        },
+                    };
+                    (id.clone(), name, value.clone())
+                }
+                _ => continue,
+            };
+
+            hints.push(Hint {
+                kind: HintKind::Value,
+                text: format!("{name} = {value}"),
```
JonasWanke: So for `a = 1`, this would create the hint `# a = 1`? Is the name in the hint really necessary?

compiler/src/language_server/hints/fuzzer.rs
```diff
@@ -0,0 +1,176 @@
...
+            .values_mut()
+            .filter(|fuzzer| matches!(fuzzer.status, Status::StillFuzzing { .. }))
+            .collect_vec();
+        log::trace!(
+            "Fuzzer running. {} fuzzers for relevant closures are running.",
+            running_fuzzers.len(),
+        );
+
+        running_fuzzers.shuffle(&mut thread_rng());
+        let fuzzer = running_fuzzers.pop()?;
```
JonasWanke: Again, can we pick a random one instead of shuffling the whole vec?

compiler/src/language_server/hints/fuzzer.rs
```diff
@@ -0,0 +1,176 @@
...
+                let id = fuzzer.closure_id.clone();
+                let first_hint = {
+                    let parameter_names = match db.find_expression(id.clone()) {
+                        Some(Expression::Lambda(Lambda { parameters, .. })) => parameters
+                            .into_iter()
+                            .map(|parameter| parameter.keys.last().unwrap().to_string())
+                            .collect_vec(),
+                        Some(_) => {
+                            log::warn!("Looks like we fuzzed a non-closure. That's weird.");
+                            continue;
```
JonasWanke: I think we can panic here

compiler/src/language_server/hints/fuzzer.rs
```diff
@@ -0,0 +1,176 @@
...
+                            log::warn!(
+                                "Using fuzzing, we found a possible error in a generated closure."
+                            );
+                            continue;
+                        }
+                    };
+                    Hint {
+                        kind: HintKind::Fuzz,
+                        text: format!(
+                            "If this is called with {},",
```
JonasWanke, line 106: We could also think about showing these values besides the parameter declarations, maybe that's easier to grasp?

compiler/src/language_server/hints/fuzzer.rs
```diff
@@ -0,0 +1,176 @@
...
+                };
+
+                let second_hint = {
+                    let panicking_inner_call = tracer
+                        .log()
+                        .iter()
+                        .rev()
+                        .filter(|entry| {
+                            let inner_call_id = match entry {
+                                TraceEntry::CallStarted { id, .. } => id,
```
JonasWanke, line 127: Now that `panic` is also implemented via `needs`, can we maybe get rid of this case?

compiler/src/language_server/hints/mod.rs
```diff
@@ -0,0 +1,198 @@
+//! Unlike the usual language server features, hints are not generated on-demand
+//! with the usual request-response model. Instead, a hints server runs in the
+//! background all the time. That way, the hints can progressively get better.
+//! For example, when opening a long file, the hints may appear from top to
+//! bottom as more code is evaluated. Then, the individual closures could get
+//! fuzzed with ever-more-complex inputs, resulting in some error cases to be
+//! displayed over time.
+//! While doing all that, we can pause regularly between executing instructions
```
JonasWanke:
```suggestion
//! displayed over time.
//!
//! While doing all that, we can pause regularly between executing instructions
```

compiler/src/language_server/hints/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+//! so that we don't occupy a single CPU at 100%.
```
JonasWanke:
```suggestion
//! so that we don't occupy a single CPU at 100 %.
```

compiler/src/language_server/hints/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+        if self.last_sent.get(&input) != Some(&hints) {
+            self.last_sent.insert(input.clone(), hints.clone());
+            self.sender.send((input, hints)).await.unwrap();
+        }
+    }
+}
+
+/// VSCode trims multiple leading spaces to one. That's why we use an
+/// [em quad](https://en.wikipedia.org/wiki/Quad_(typography)) instead, which
+/// per definition has the same width as a normal space.
```
JonasWanke: That's not quite its definition: “An em quad is a space that is one [em](https://en.wikipedia.org/wiki/Em_(typography)) wide; as wide as the height of the font.” In fact, that is usually wider than one space, as is also visualized on this handy page: https://jkorpela.fi/chars/spaces.html. For monospace fonts it might work, but not in general.

compiler/src/language_server/hints/utils.rs
```diff
@@ -0,0 +1,26 @@
...
+        .line;
+    let line_start_offsets = db.line_start_utf8_byte_offsets(id.input.clone());
+    let last_characer_of_line = if line as usize == line_start_offsets.len() - 1 {
+        db.get_input(id.input.clone()).unwrap().len()
+    } else {
+        line_start_offsets[(line + 1) as usize] - 1
+    };
+    let position = db
+        .offset_to_lsp(id.input.clone(), last_characer_of_line)
+        .to_position();
```
JonasWanke:
```suggestion
    let last_character_of_line = if line as usize == line_start_offsets.len() - 1 {
        db.get_input(id.input.clone()).unwrap().len()
    } else {
        line_start_offsets[(line + 1) as usize] - 1
    };
    let position = db
        .offset_to_lsp(id.input.clone(), last_character_of_line)
        .to_position();
```

compiler/src/language_server/references.rs
```diff
@@ -220,16 +220,17 @@ impl<'a> Context<'a> {
...
-            range: lsp_types::Range {
-                start: self
-                    .db
-                    .offset_to_lsp(id.input.clone(), span.start)
-                    .to_position(),
-                end: self.db.offset_to_lsp(id.input, span.end).to_position(),
-            },
-            kind: Some(kind),
-        });
+        if let Some(span) = self.db.hir_id_to_span(id.clone()) {
```
JonasWanke, line 223: Do you have an example where this failed?

JonasWanke (changes requested):

compiler/src/fuzzer/mod.rs
MarcelGarus: Not quite because if it's fuzzing some input and the function returns normally after 90 instructions, we stop. With the for-loop, we would then try again with a different input etc.

Should we change the name of the function or the behavior of the function to continue fuzzing with a reduced number of instructions left?

compiler/src/language_server/hints/fuzzer.rs
MarcelGarus: Maybe, not sure.

I propose we leave it like this for now, but keep that in the back of our heads. I think after we implemented all hints we want to show, it generally makes sense to brainstorm about how we can improve their presentation.

compiler/src/fuzzer/mod.rs
JonasWanke: Yes, I think that would make sense. I lean towards the latter, though we could also let it return how many instructions were actually executed or how many are left

compiler/src/language_server/references.rs
MarcelGarus: The server crashed when I hovered over the `use` function 😅

compiler/src/fuzzer/mod.rs
MarcelGarus: Okay, I implemented that. I also thought about this pattern of giving in the number of instructions, which we now use both for the VM directly as well as the fuzzer.

In the long term, instead of providing the number of instructions to run, it could make sense to instead provide a closure that decides whether to keep going. The closure could be called before every instruction.
That would allow us to implement several abortion strategies, e.g. aborting after a certain wall time or after a number of instructions or as our CPU usage rises above some threshold, all without requiring us to keep track of the concrete metric in the code.

compiler/src/language_server/hints/fuzzer.rs
```diff
@@ -124,13 +121,12 @@ impl FuzzerManager {
                         .iter()
                         .rev()
                         .filter(|entry| {
-                            let inner_call_id = match entry {
-                                TraceEntry::CallStarted { id, .. } => id,
```
JonasWanke: You can also remove lines 137 – 139 now

compiler/src/fuzzer/fuzzer.rs
```diff
@@ -62,37 +51,66 @@ impl Fuzzer {
...
+    }
+
+    pub fn run(&mut self, db: &Database, mut num_instructions: usize) {
+        let mut status = mem::replace(&mut self.status, None).unwrap();
+        while matches!(status, Status::StillFuzzing { .. }) {
+            let (new_status, num_instructions_executed) =
+                self.map_status(db, status, num_instructions);
+            status = new_status;
+
+            if num_instructions_executed > num_instructions {
```
JonasWanke:
```suggestion
            if num_instructions_executed >= num_instructions {
```

JonasWanke (changes requested):

compiler/src/language_server/hints/fuzzer.rs
MarcelGarus: Oops, it not that easy after all.

JonasWanke (approved):

PR #231, opened 2022-07-24
Change panic/needs message value to reason string

MarcelGarus:
Depends on #230

### DIFF
--- a/compiler/src/compiler/lir.rs
+++ b/compiler/src/compiler/lir.rs
@@ -63,7 +63,7 @@ pub enum Instruction {
     },
 
     /// Pops a boolean condition and a reason. If the condition is true, it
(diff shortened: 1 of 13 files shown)


compiler/src/compiler/lir.rs
```diff
@@ -62,11 +62,11 @@ pub enum Instruction {
         num_args: usize,
     },
 
-    /// Pops a boolean condition and a message. If the condition is true, it
+    /// Pops a boolean condition and a reason. If the condition is true, it
     /// just pushes Nothing. If the condition is false, it panic with the
```
JonasWanke:
```suggestion
    /// just pushes Nothing. If the condition is false, it panics with the
```

JonasWanke (changes requested):

JonasWanke (approved):

PR #232, opened 2022-07-26
Update language documentation

MarcelGarus:
Although this is far from complete, I think we can already merge this.
It's definitely more comprehensive and confusing than having several semi-outdated idea-dumps scattered throughout the project.


language.md
```diff
@@ -0,0 +1,543 @@
...
+- [Needs](#needs)
+- [Pattern matching](#pattern-matching)
+- [Concurrency](#concurrency)
+  - [Channels](#channels)
+- [Packages](#packages)
+- [Environment](#environment)
+- [Interoperability with other languages](#interoperability-with-other-languages)
+  - [Add to the Environment](#add-to-the-environment)
+  - [Contain Pure Code](#contain-pure-code)
+- [Deploying code](#deploying-code)
```
JonasWanke: Some headers use title case, others don't

language.md
```diff
@@ -0,0 +1,543 @@
...
+
+## Basic Syntax
+
+Candy's syntax is inspired by Elm and Haskell.
+
+Source code is stored in plain text files with a `.candy` file extension.
+
+Line [comments](#comments) start with `#` and end at the end of the line:
```
JonasWanke: As we only have line comments and it's explained that they go to the end of the line, I don't think we have to write “Line comments”

```suggestion
[Comments](#comments) start with `#` and end at the end of the line:
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+
+```candy
+# This is a comment.
+```
+
+Naming rules are similar to other programming languages.
+Identifiers start with a lowecase letter and may contain letters or digits.
```
JonasWanke:
```suggestion
Identifiers start with a lowercase letter and may contain letters or digits.
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+  baz = 3
+  5
+# bar is now 5
+```
+
+## Values
+
+All data in your program is composed of values.
+Values can be created through literals, expressions that evaluate to a value.
+All values are immutable – once created, they do not change.
```
JonasWanke:
```suggestion
All values are immutable – once created, they do not change.
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+
+### Texts
+
+Texts are Unicode strings.
+They start and end with double quotes (`"`) and can span multiple lines if they're indented.
+
+You can also start texts with any number of single quotes (`'`) followed by a double quote (`"`).
+This so-called meta-text can only be ended with a double quote and the same number of single quotes that it started with.
+
+Finally, you can use curly braces (`{}`) containing a value to insert a stringified version of the value into the string at that position.
```
JonasWanke: I am thinking about only allowing strings to be inserted, meaning you have to convert your value to a string before inserting it in an interpolation. The reason is that the debug-formatted string is not always what you want and it can be easy to forget to call your custom stringification method if inserting the raw value also works. We could also think about one special character in interpolations as a shortcut for the debug string, though I don't want to add format strings with more features.

language.md
```diff
@@ -0,0 +1,543 @@
...
+
+```candy
+"Hello!"
+"A somewhat
+  long
+  text."
+'"This is a meta text, where you can use " inside the text."'
+''"This is a double-meta text, allowing you to use "' inside it without ending it."''
+"Some {interpolation}."
+'"In meta texts, {{interpolation}} requires more curly braces, otherwise the values are {not interpolated}."'
```
JonasWanke:
```suggestion
'"In meta texts, {{interpolation}} requires more curly braces; otherwise, the values are {not interpolated}."'
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+
+```candy
+foo =
+  bar = hello  # error because hello doesn't exist yet
+  hello = 5
+  4
+bar = hello  # error because hello doesn't exist anymore
+```
+
+Variables at the top level of a file are visible to the module system.
```
JonasWanke:
```suggestion
Variables at the file's top level are visible to the module system (“public”) if they are declared using `:=`.
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+All other variables are local.
+
+Declaring a variable with the same name as another simply shadows that variable:
```
JonasWanke:
```suggestion
Declaring a variable with the same name as another simply shadows that variable, though that's not allowed for public variables.
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+foo = 3
+# there is no way to get to the 5
+bar =
+  foo = 4
+# here, foo is still 3
+```
+
+## Functions
+
+Functions can either be defined using closure literals or by writing them as parameterized variables with arguments in front of the `=`.
```
JonasWanke:
```suggestion
Functions can either be defined using closure literals or by writing them as parameterized variables with arguments in front of the `=` or `:=`.
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+Both representations are equivalent with respect to what they do during runtime.
+
+```candy
+# Both these definitions are equivalent.
```
JonasWanke: Already stated above about their behavior during runtime. During compile-time, however, they differ with regards to fuzzing.

language.md
```diff
@@ -0,0 +1,543 @@
...
+# inside red/yellow/.candy
+
+foo = use ".purple"  # imports the purple child module
+foo = use "..blue"   # imports the blue sibling module
+foo = use "...green" # imports the green parent module
+```
+
+Each additional dot at the beginning symbolizes a navigation one level up.
+The possible multiple dots are followed by the name of the module to import.
+Note that you can't navigate further than one level in – for example, the `yellow` module can't import the `brown` module, only its parent module `green`.
```
JonasWanke:
```suggestion
Note that you can't navigate further than one level in – for example, the `yellow` module can't import the `brown` module, only its parent module `green`.
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+brown =
+  foo = 3
+  bar = foo
+  baz a = a
+  [
+    Bar: foo,
+    Baz: bar,
+  ]
+
+# equivalent:
```
JonasWanke: Except for the fuzzing behavior

language.md
```diff
@@ -0,0 +1,543 @@
...
+brown = [
+  Bar: 5,
+  Baz: { a -> a },
+]
+```
+
+The `useAsset` also allows you to import arbitrary non-Candy-files part from your module hierarchy.
```
JonasWanke:
```suggestion
The `useAsset` also allows you to import arbitrary non-Candy files that are part of your module hierarchy.
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+
+```
+main.candy
+translations.json
+```
+
+```
+# inside main.candy
+
+translations = json.parse(useAsset "..translations.json")
```
JonasWanke:
```suggestion
translations = json.parse (useAsset "..translations.json")
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+## Comments
+
+TODO: Write something including doc comments
+
+## Panics
+
+Candy programs can panic, causing them to crash.
+Contrary to crashes in other programming languages, it's always programmatically clear which part of the code is at fault.
+
+For example, in Rust, if a function panics you have to look at its documentation to understand if the panic is your fault or not:
```
JonasWanke:
```suggestion
For example, in Rust, if a function panics, you have to look at its documentation to understand if the panic is your fault or not:
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+To enforce structured concurrency, they can only be spawned using a special concurrency object called the *nursery*.
+In the following code, the surrounding call to `core.parallel` only exits when all fibers inside have completed.
+
+```candy
+core.parallel { nursery ->
+  core.fiber.spawn nursery { print "Banana" }
+  core.fiber.spawn nursery { print "Kiwi" }
+  # Banana and Kiwi may print in any order
+}
+print "Peach"  # Always prints after the others
```
JonasWanke, line 469: I don't think this simple `print` call would work with capabilities

language.md
```diff
@@ -0,0 +1,543 @@
...
+
+TODO: Write about async await as soon as they are in the Core package
+
+## Packages
+
+TODO: Write something
+
+## Environment
+
+At some point, your Candy program needs to have side-effects – otherwise, it's just heating up your CPU.
```
JonasWanke:
```suggestion
At some point, your Candy program needs to have side effects – otherwise, it's just heating up your CPU.
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+You can put the resulting WASM module in your Candy module hierarchy, call `useAsset` with that file, and pass the data to a WASM runtime that we'll write in Candy when we get to it.
+
+This approach moves the native code entirely into the Candy domain, so the Candy compiler can also reason about the WASM code.
+For example, if you call a function of your WASM module only with inputs known at compile-time, the Candy compiler may execute those calls directly and not include the original WASM code in the binary at all.
+
+## Deploying code
+
+This chapter is especially experimental and spitball-y.
+
+Similar to how Zig build scripts work, we may support having a `build.candy` file that contains information about how to compile and optimize your code.
```
JonasWanke, line 578: What would be written in there? Should this really be something customizable?

language.md
```diff
@@ -0,0 +1,543 @@
...
+
+Next to the interpreted VM, we plan to compile to LLVM or WASM.
+
+One idea we had is to let you provide a custom scoring function in the build script instead of having binary options like "optimize for speed" or "optimize for performance".
+This scoring function could get used by the optimizer to choose which paths to take.
+For example, you could formulate a build where you're okay with a resulting binary blowup of 1 KiB per 10 ms of saved time in some annotated performance-crititcal section.
```
JonasWanke:
```suggestion
For example, you could formulate a build where you're okay with a resulting binary blowup of 1  KiB per 10 ms of saved time in some annotated performance-critical section.
```

language.md
```diff
@@ -0,0 +1,543 @@
...
+Or when developing for an embedded device with limited storage capacity, you might want to generate a binary that fits in the limit but is otherwise as fast as possible.
```
JonasWanke:
```suggestion
Or, when developing for an embedded device with limited storage capacity, you might want to generate a binary that fits within the limit but is otherwise as fast as possible.
```

JonasWanke (changes requested):

language.md
```diff
@@ -158,19 +158,21 @@ longClosure = { foo ->
...
+TODO: Or are they called ports?
+
 Channels ends allow you to interact with a [channel](#concurrency).
 There are receive ends and send ends to receive and send data from a channel, respectively.
 
+### More?
+
+TODO: Lists? Sets?
+
+- lists: `[1, 2, 3]`
```
JonasWanke: Were just merged :)

JonasWanke (changes requested):

language.md
```diff
@@ -140,10 +140,28 @@ Structs are mappings from keys to values (also known as dictionaries or hash map
...
   Name: "Candy",
   Foo: 42,
+  "TextKey": 4,
+  3: 2,
 ]
 ```
 
-TODO: Struct access using dot
+To use numbers as keys, you can also omit the keys of 
+
```
JonasWanke:
```suggestion
```

language.md
```diff
@@ -140,10 +140,28 @@ Structs are mappings from keys to values (also known as dictionaries or hash map
...
+Omitting keys is equivalent to using zero-indexed numbers as keys:
+
+```candy
+[Hello, World]
+
+# equivalent:
+[0: Hello, 1: World]
```
JonasWanke:
```suggestion
Omitting keys is equivalent to using zero-indexed numbers as keys. You can also add fields with explicit keys after these:

```candy
[Hello, World, Foo: Bar]

# equivalent:
[0: Hello, 1: World, Foo: Bar]
```

language.md
```diff
@@ -411,9 +428,31 @@ mySqrt a =               # If you pass `a = -1`,
 
 ## Pattern Matching
 
-TODO: Write something
+Candy supports structural pattern matching using the match operator `%`.
+
+```candy
+match (foo 5) %
```
JonasWanke: This looks like `maybe` is also part of the necessary syntax. Maybe call a different function?

language.md
```diff
@@ -411,9 +428,31 @@ mySqrt a =               # If you pass `a = -1`,
...
+  [Ok, value] -> ...
+  [Error, errorValue], core.int.isEven errorValue -> ...
+  _ -> ...
+```
+
+Here, each indented line after the match operator represents a match case.
+Each case can match based on the pattern as well an optional condition separated by a comma.
+The first matching case is executed.
+If no case matches, you code panics.
```
JonasWanke:
```suggestion
If no case matches, your code panics.
```

language.md
```diff
@@ -411,9 +428,31 @@ mySqrt a =               # If you pass `a = -1`,
...
 
-TODO: Irrefutable and refutable patterns
+If you're sure about the structure of a value, you can also use patterns on the left-hand side of an assignment.
+These are called irrefutable patterns.
+Again, if the pattern doesn't match, the code panics.
+
+```candy
+[a, b] = myStruct
```
JonasWanke:
```suggestion
[a, b] = myList
```

JonasWanke (changes requested):

JonasWanke (approved):

PR #233, opened 2022-07-26
Extend Core

JonasWanke:
Some code for testing:

```candy
# Int
result = core.int.add 1234567890123456789012345678901234567890 1
result = core.int.subtract 5 7
result = core.int.negate 42
result = core.int.multiply 5 7
result = core.int.divideTruncating 42 5
result = core.int.modulo 42 5

result = core.int.compareTo 7 5
result = core.int.isLessThan 7 5
result = core.int.isLessThanOrEqualTo 7 5
result = core.int.isGreaterThan 7 5
result = core.int.isGreaterThanOrEqualTo 7 5

result = core.int.shiftLeft 9 2
result = core.int.shiftRight 9 2
result = core.int.shiftLeft (negate 9) 2
result = core.int.shiftRight (negate 9) 2

result = core.int.bool.is True
result = core.int.bool.is False

result = core.int.min 7 5
result = core.int.max 7 5
result = core.int.coerceAtLeast 7 5
result = core.int.coerceAtMost 7 5
result = core.int.coerceIn 1 7 5

result = core.int.parse "123"
result = core.int.parse "-456"
result = core.int.parse "Candy"

# Result
result = core.result.ok 123
result = core.result.error 123
result = core.result.is (ok 123)
result = core.result.is (error 123)
```

Major remaining TODOs compared to the previous Core library: collections, IO, and ranges.

Some methods from Dart, Rust, and previous Candy versions I thought about but haven't yet included:

* Bool
  ```candy
  then condition body =
    needs (is condition)
    needs (function.is0 body)
    ✨.ifElse condition body { False }
  ```
* Conditionals
  * if (without else) could return a result for chaining
  * loop
* Int
  * int.times
  * sign
  * toRadixString(int radix) → String
  * toString() → String
  * toStringAsExponential([int? fractionDigits]) → String
  * toSigned(int width) → int https://api.dart.dev/stable/2.17.6/dart-core/int/toSigned.html
  * toUnsigned(int width) → int https://api.dart.dev/stable/2.17.6/dart-core/int/toUnsigned.html
* Result
  * is_err_and
  * is_ok_and
  * iter
* Text
  * char_indices
  * compareTo
  * eq_ignore_ascii_case
  * find, rfind, indexOf, lastIndexOf
  * get
  * into_string
  * is_ascii
  * is_char_boundary
  * lines
  * make_ascii_lowercase, make_ascii_uppercase
  * match_indices
  * matches
  * replace
  * replaceAll, replaceAllMapped
  * replaceFirst, replaceFirstMapped
  * replacen
  * replaceRange
  * rmatch_indices
  * rmatches
  * rsplit
  * rsplit_once
  * rsplit_terminator
  * rsplitn
  * split
  * split_ascii_whitespace
  * split_at
  * split_inclusive
  * split_once
  * split_terminator
  * split_whitespace
  * splitn
  * to_ascii_lowercase, to_ascii_uppercase
  * to_lowercase, to_uppercase
  * trim_matches, trim_start_matches, trim_end_matches

  ```candy
  padLeft text targetWidth padding :=
    needs (is text)
    needs (int.is targetWidth)
    needs (is padding)
    needs (equals (length padding) 1)
    # TODO
  padRight text targetWidth padding :=
    needs (is text)
    needs (int.is targetWidth)
    needs (is padding)
    needs (equals (length padding) 1)
    # TODO

  repeat text count :=
    needs (is text)
    needs (int.is count)
    needs (int.isNonNegative count)
    # TODO

  split text pattern :=
    needs (is text)
    needs (is pattern)
    needs (bool.not (isEmpty pattern))
    ✨.textSplit text pattern
  splitFirst text pattern limit :=
    needs (is text)
    needs (is pattern)
    needs (bool.not (isEmpty pattern))
    ✨.textSplit text pattern
  ```

I also just noticed that we now have `run` and `call`. I suggest we either leave `call` and `doNotCall` or `run` and `doNotRun` and move both to the functions module.

### Checklist


compiler/src/builtin_functions.rs
```diff
@@ -4,17 +4,40 @@ use strum_macros::EnumIter;
...
+    GetArgumentCount,    // closure -> argumentCount
+    IfElse,              // condition thenClosure elseClosure -> resultOfExecutedClosure
+    IntAdd,              // (summandA: int) (summandB: int) -> (sum: int)
+    IntBitLength,        // (value: int) -> (numberOfBits: int)
+    IntBitwiseAnd,       // (valueA: int) (valueB: int) -> (result: int)
+    IntBitwiseOr,        // (valueA: int) (valueB: int) -> (result: int)
+    IntBitwiseXor,       // (valueA: int) (valueB: int) -> (result: int)
+    IntCompareTo,        // (valueA: int) (valueB: int) -> (ordering: Less | Equal | Greater)
+    IntDivideTruncating, // (dividend: int) (divisor: int) -> (quotient: int)
+    IntModulo,           // (dividend: int) (divisor: int) -> (remainder: int)
```
MarcelGarus: In some languages (such as Zig), there's a difference between `modulo` and `remainder`, where modulo for negative numbers may return a negative number, but remainder always returns a non-negative one (or was it the other way around?).

Which one is used here? (And just mentioning this because the return value is called remainder here.)

compiler/src/builtin_functions.rs
```diff
@@ -4,17 +4,40 @@ use strum_macros::EnumIter;
...
+    IntMultiply,         // (factorA: int) (factorB: int) -> (product: int)
+    IntParse,            // (text: text) -> (parsedInt: maybeOfInt)
```
MarcelGarus: I'm not entirely sure when the arguments in the comments have types and when not, but this looks like a very obvious candidate for omitting the type.

```suggestion
    IntParse,            // text -> (parsedInt: maybeOfInt)
```

compiler/src/compiler/rcst.rs
```diff
@@ -410,10 +412,15 @@ impl<T: SplitOuterTrailingWhitespace> SplitOuterTrailingWhitespace for Option<T>
...
     fn split_outer_trailing_whitespace(self) -> (Vec<Rcst>, Self) {
-        let (whitespace, second) = self.1.split_outer_trailing_whitespace();
-        (whitespace, (self.0, second))
+        if self.1.is_empty() {
+            let (whitespace, first) = self.0.split_outer_trailing_whitespace();
+            (whitespace, (first, self.1))
+        } else {
+            let (whitespace, second) = self.1.split_outer_trailing_whitespace();
+            (whitespace, (self.0, second))
+        }
```
MarcelGarus: Maybe something like this is more readable?

```suggestion
        let (left, right) = self;
        if right.is_empty() {
            let (whitespace, first) = left.split_outer_trailing_whitespace();
            (whitespace, (first, right))
        } else {
            let (whitespace, second) = right.split_outer_trailing_whitespace();
            (whitespace, (left, second))
        }
```

compiler/src/compiler/string_to_rcst.rs
```diff
@@ -300,7 +303,7 @@ mod parse {
             Some((
                 " ",
                 Rcst::Int {
-                    value: 42,
+                    value: BigUint::from(42u8),
```
MarcelGarus: Because `BigUint` implements `From<i64>` (and the same for several types of `uXX`), I think we might be able to write something like this:

```suggestion
                    value: 42.into(),
```

This also applies to the other places where the `BigUint::from(constant)` pattern is used.

compiler/src/compiler/cst_to_ast.rs
```diff
@@ -275,7 +280,7 @@ impl LoweringContext {
                 let mut errors = vec![];
 
                 assert!(
-                    !matches!(opening_bracket.kind, CstKind::OpeningBracket),
+                    matches!(opening_bracket.kind, CstKind::OpeningBracket),
```
MarcelGarus, line 283: Interesting, I'm wondering how this succeeded before. Maybe, the `opening_bracket` is a `TrailingWhitespace` containing the actual `OpeningBracket`.

packages/Core/Bool.candy
```diff
@@ -1,15 +1,55 @@
+equals = (use "..Equality").equals
+function = use "..Function"
 type = use "..Type"
 
-is value := type.is value Bool
+is value :=
+  ✨.ifElse
+    (✨.equals value True)
```
MarcelGarus: I believe we don't even need the parentheses here because the outer arguments are separated using indentation.

```suggestion
    ✨.equals value True
```

Not sure though 😅

packages/Core/ControlFlow.candy
```diff
@@ -0,0 +1,40 @@
...
+  needs (isFunction0 then)
+  ifElse condition then {}
+
+run body =
+  # A function that runs the given `body` with no arguments and returns its result.
+  #
+  # ```
+  # let a = run {
+  #   // Some code that can create local variables without cluttering the surrounding namespace.
+  # }
```
MarcelGarus:
```suggestion
  # a = run {
  #   ## Some code that can create local variables without cluttering the surrounding namespace.
  # }
```

packages/Core/Function.candy
```diff
@@ -7,9 +7,13 @@ getArgumentCount function :=
...
+is0 value := ✨.ifElse (is value) { equals (getArgumentCount value) 0 } { False }
+is1 value := ✨.ifElse (is value) { equals (getArgumentCount value) 1 } { False }
+is2 value := ✨.ifElse (is value) { equals (getArgumentCount value) 2 } { False }
+is3 value := ✨.ifElse (is value) { equals (getArgumentCount value) 3 } { False }
+is4 value := ✨.ifElse (is value) { equals (getArgumentCount value) 4 } { False }
+is5 value := ✨.ifElse (is value) { equals (getArgumentCount value) 5 } { False }
+
+call function0 :=
+  needs (is0 function0)
+  ✨.call function0
```
MarcelGarus: The core package now exports both `call` and `run`, which essentially do the same thing. Is that intentional? Or do we want to reduce this to one function?

packages/Core/Int.candy
```diff
@@ -1,8 +1,151 @@
...
+modulo dividend divisor :=
+  needs (is dividend)
+  needs (is divisor)
+  ✨.intModulo dividend divisor
+
+compareTo valueA valueB :=
+  needs (is valueA)
+  needs (is valueB)
+  result = ✨.intCompareTo valueA valueB
+  bool.implies (equals result Equal) (equals valueA valueB)
```
MarcelGarus: Why is this needed? Or is it supposed to just be a sanity check?

```suggestion
  check (bool.implies (equals result Equal) (equals valueA valueB))
```

MarcelGarus (changes requested):

packages/Core/Function.candy
JonasWanke: Have a look at the last sentence in the (pretty long) PR description, right above the checklist we never use. I want to remove one, but am not sure which one we want to keep

compiler/src/builtin_functions.rs
JonasWanke: Ah, this uses the `%` operator which in turn uses the `Rem` trait. For standard number types, the impl contains the comment “This operation satisfies `n % d == n - (n / d) * d`. The result has the same sign as the left operand.” For `BigInt`, it uses [`div_rem(…)`](https://docs.rs/num-integer/0.1.45/num_integer/trait.Integer.html#tymethod.div_rem). I'll fix the naming in our code. Do we want to provide both operations or only one (which?)?

packages/Core/Int.candy
JonasWanke: Just a sanity check

packages/Core/Bool.candy
JonasWanke: Makes sense and the parser doesn't complain. I've simplified this in the remaining code as well

compiler/src/builtin_functions.rs
MarcelGarus: That's a good question. I think it doesn't hurt to provide both. (And both operations exist, so I guess there are use cases for both of them.)

packages/Core/Function.candy
MarcelGarus: Ahh well 😅 I prefer `run` because it better describes the longer-term intention. Like, of course, you _call_ a function, but only in order to _run_ it (and to get the return value, but the side effects of the function execution are probably the most important part here, otherwise the function could be a constant).

packages/Core/Int.candy
MarcelGarus: In that case, we'd need the `check` as suggested above.

MarcelGarus (approved):

PR #234, opened 2022-07-29
List literals

JonasWanke:
**This depends on #233!**

Structs can now contain positional fields in addition to named fields: `[Foo, Bar: Baz]` is equivalent to `[0: Foo, Bar: Baz]`. Positional fields must be written before any named fields, and they are indexed sequentially, starting with zero.


compiler/src/compiler/rcst.rs
```diff
@@ -180,13 +179,14 @@ impl Display for Rcst {
...
             Rcst::StructField {
-                key,
-                colon,
+                key_and_colon,
                 value,
                 comma,
             } => {
-                key.fmt(f)?;
-                colon.fmt(f)?;
+                if let Some(box (key, colon)) = key_and_colon {
```
MarcelGarus, line 186: Whoa that `box` is fancy!

compiler/src/compiler/ast.rs
```diff
@@ -154,15 +156,15 @@ impl FindAst for Ast {
...
-        None
+        self.positional_fields
+            .iter()
+            .filter_map(|value| value.find(id))
+            .chain(
+                self.named_fields
+                    .iter()
+                    .filter_map(|(key, value)| key.find(id).or_else(|| value.find(id))),
+            )
+            .next()
```
MarcelGarus: We can simplify this by first putting all values into an iterator and then trying to find the ID on that.

```suggestion
        self.positional_fields
            .iter()
            .chain(self.named_fields.iter().flat_map(|(key, value)| vec![key, value].iter()))
            .filter_map(|value| value.find(id))
            .next()
```

Not entirely sure about the ownership of the `vec` though

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #236, opened 2022-07-29
Add compiler CI

JonasWanke:
There are a few Clippy complaints left:

* AST's `CallReceiver::Call` is not yet created while parsing, but is already handled in the HIR, hence I didn't want to remove it. The parser handling can be done separately.
* Some modules (`fuzzer`, `vm`) contain modules with the same name – do we want to give them a different name or ignore this lint?
* Two `to_*` methods are not idiomatic with respect to using a reference or consuming the value.
* and a few unused methods

### DIFF
--- a/compiler/src/compiler/cst_to_ast.rs
+++ b/compiler/src/compiler/cst_to_ast.rs
@@ -53,8 +53,7 @@ fn ast(db: &dyn CstToAst, input: Input) -> Option<(Arc<Vec<Ast>>, HashMap<ast::I
     let asts = match db.cst(input.clone()) {
         Ok(cst) => {
             let cst = cst.unwrap_whitespace_and_comment();
-            let asts = (&mut context).lower_csts(&cst);
-            asts
+            context.lower_csts(&cst)
         }
         Err(InvalidInputError::DoesNotExist) => return None,
         Err(InvalidInputError::InvalidUtf8) => {
@@ -144,7 +143,7 @@ impl LoweringContext {
                 );
 
                 let text = parts
-                    .into_iter()
+                    .iter()
                     .filter_map(|it| match it {
                         Cst {
                             kind: CstKind::TextPart(text),
@@ -281,7 +280,7 @@ impl LoweringContext {
                 );
 
                 let fields = fields
-                    .into_iter()
+                    .iter()
                     .filter_map(|field| {
                         if let CstKind::StructField {
                             key,
@@ -493,6 +492,7 @@ impl LoweringContext {
             CstKind::Identifier(identifier) => {
                 self.create_string(key.id.to_owned(), identifier.uppercase_first_letter())
             }
+            // TODO: handle CstKind::Error
             _ => panic!(
                 "Expected an identifier after the dot in a struct access, but found `{}`.",
                 key
@@ -508,7 +508,7 @@ impl LoweringContext {
     fn lower_parameters(&mut self, csts: &[Cst]) -> (Vec<AstString>, Vec<CompilerError>) {
         let mut errors = vec![];
         let parameters = csts
-            .into_iter()
+            .iter()
             .enumerate()
             .map(|(index, it)| match self.lower_parameter(it) {
                 Ok(parameter) => parameter,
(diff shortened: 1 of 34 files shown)


compiler/src/compiler/string_to_rcst.rs
```diff
@@ -80,16 +80,12 @@ mod parse {
...
+    static SUPPORTED_WHITESPACE: &str = " \r\n\t";
 
     fn literal<'a>(input: &'a str, literal: &'static str) -> Option<&'a str> {
         log::trace!("literal({input:?}, {literal:?})");
-        if input.starts_with(literal) {
-            Some(&input[literal.len()..])
-        } else {
-            None
-        }
+        input.strip_prefix(literal)
```
MarcelGarus, line 88: Whoa, was that detected by clippy? Impressive!

MarcelGarus (approved):

MarcelGarus:
About the reported lints:

- Regarding `CallReceiver`: What exactly is the benefit of analyzing a call's receiver directly in the AST over just storing an arbitrary `Ast` expression as the receiver? Like, sure, we'll find out in the AST that the call is invalid if the receiver is an int literal or something like that, but we would also find those cases during constant evaluation. So, in terms of developer experience, it's not really a difference.
  What's more, in the next stage of the compiler pipeline, the AST-to-HIR conversion, we don't actually use the receiver information in a meaningful way: After special-casing the calling of a  `AstIdentifier::("needs")`, we just match over the receiver type, and if it's a `StructAccess`, we call `lower_struct_access`, if it's a `Call`, we call `lower_call`, and if it's an `Identifier`, we essentially have a copied version of the `compile_single`' code for lowering an `Identifier`. We could just as easily treat all receivers the same (just `compile_single` them directly).
- Regarding modules with the same name: I think it makes sense to ignore this lint. For example, I like that modules can have local `utils` modules without polluting some outer scope.
- Regarding the `to_*` methods: I guess it makes sense to rename them then.
- Few unused methods: That's probably okay for now. Some of them, such as `Hir::return_value`, at least feel like they could become useful later on. But in the long term, ofc it makes sense to clean these up too.

JonasWanke:
Regarding the modules with the same name: I think my explanation wasn't clear, so here's a link to the lint: https://rust-lang.github.io/rust-clippy/master/index.html#module_inception. The problem is module `fuzzer` _inside_ module `fuzzer`

compiler/src/compiler/cst_to_ast.rs
```diff
@@ -233,12 +232,12 @@ impl LoweringContext {
...
-                        CallReceiver::StructAccess(self.lower_struct_access(struct_, dot, key)),
-                    ),
+                    CstKind::Identifier(identifier) => {
+                        Some(self.lower_identifier(name.id, identifier.to_string()))
+                    }
+                    CstKind::StructAccess { struct_, dot, key } => {
+                        Some(self.lower_struct_access(name.id.to_owned(), struct_, dot, key))
+                    }
                     _ => None,
                 };
```
MarcelGarus: Doesn't the same reasoning that motivated the removal of a `CallReceiver` also apply here? i.e. not special-casing based on the type of the receiver and just doing `let receiver = self.lower_cst(name);` That would also allow for `lower_identifier` and `lower_struct_access` to be inlined.

Sidenote: It probably also makes sense to rename `CstKind::Call`'s `name` to `receiver`.

MarcelGarus (changes requested):

MarcelGarus:
Regarding the `module_inception` lint: That's a fair concern, I agree that having both a `mod.rs` and `fuzzer.rs` in `fuzzer` is kinda weird. We could merge both into the `mod.rs` file, but they _do_ contain different logic (the CLI and exports vs. the actual fuzzing logic).

In the best case, I think we'd rename the `fuzzer.rs` to something else, but I'm having a hard time finding a fitting name. We could try to go for a name that describes the module's role _within_ the fuzzer, i.e. `engine` or something like that. What do you think?

MarcelGarus (approved):

MarcelGarus:
Just noticed that the term `receiver` probably stems from object-oriented programming languages, where method invocations are implemented using message passing and a better term in Candy would probably be `callee`. But we can change that later.

PR #237, opened 2022-08-03
Replace `Input` with `Module`

MarcelGarus:
This PR has the following changes:

- `Input` is now called `Module`
- `Module` corresponds to semantic Candy modules instead of concrete files.
- As a consequence, `use` now works differently. In particular, `useAsset` has been merged into `use`, and it's implemented using a new instruction instead of being a very special builtin function.


compiler/src/compiler/ast_to_hir.rs
```diff
@@ -469,141 +468,25 @@ impl<'a> Context<'a> {
...
-        //     HirId(~:test.candy:use:key): HirId(~:test.candy:use:rawPath),
-        //   ]
-        //   HirId(~:test.candy:use:useLocalModule) = builtinUseLocalModule
-        //   HirId(~:test.candy:use:importedModule) = call HirId(~:test.candy:use:useLocalModule) with these arguments:
-        //     HirId(~:test.candy:use:currentPath)
-        //     HirId(~:test.candy:use:target)
+        // HirId(~:test.candy:use) = lambda { HirId(~:test.candy:use:relativePath) ->
+        //   HirId(~:test.candy:useAsset:importedFileContent) = useAssetModule
+        //     currently in ~:test.candy:useAsset:importedFileContent
+        //     relative path: HirId(~:test.candy:useAsset:relativePath)
```
JonasWanke:
```suggestion
        //   HirId(~:test.candy:use:importedFileContent) = useAssetModule
        //     currently in ~:test.candy:use:importedFileContent
        //     relative path: HirId(~:test.candy:use:relativePath)
```

compiler/src/module.rs
```diff
@@ -0,0 +1,295 @@
...
+        }
+    }
+
+    pub fn from_package_root_and_url(package_root: PathBuf, url: Url, kind: ModuleKind) -> Self {
+        match url.scheme() {
+            "file" => {
+                Module::from_package_root_and_file(package_root, url.to_file_path().unwrap(), kind)
+            }
+            "untitled" => Module::from_anonymous_content(
+                url.to_string()["untitled:".len()..].to_owned(),
```
JonasWanke:
```suggestion
                url.to_string().strip_prefix("untitled:").unwrap().to_owned(),
```

But I think this should actually be the file's content and not its URI

compiler/src/module.rs
```diff
@@ -0,0 +1,295 @@
...
+            _ => panic!("Unsupported URI scheme: {}", url.scheme()),
+        }
+    }
+    pub fn from_package_root_and_file(
+        package_root: PathBuf,
+        file: PathBuf,
+        kind: ModuleKind,
+    ) -> Self {
+        let relative_path =
+            fs::canonicalize(&file).expect("Package root does not exist or is invalid.");
```
JonasWanke:
```suggestion
            fs::canonicalize(&file).expect("File does not exist or its path is invalid.");
```

compiler/src/module.rs
```diff
@@ -0,0 +1,295 @@
...
+        .map(|it| Arc::new(it))
+}
+
+fn get_module_content(db: &dyn ModuleDb, module: Module) -> Option<Arc<Vec<u8>>> {
+    if let Some(content) = db.get_open_module_content(module.clone()) {
+        return Some(content);
+    };
+
+    if let Package::Anonymous { root_content } = module.package {
+        return Some(Arc::new(root_content.clone().into_bytes()));
```
JonasWanke: Maybe we should just store the bytes directly, potentially in an `Arc`?

compiler/src/module.rs
```diff
@@ -0,0 +1,295 @@
...
+                    let mut path = path.clone();
+                    path.set_extension("candy");
+                    path
+                },
+            ],
+        })
+    }
+
+    pub fn associated_debug_file(&self, debug_type: &str) -> PathBuf {
+        let mut path = self.to_possible_paths().unwrap().pop().unwrap();
```
JonasWanke: This `unwrap()` will panic when trying to save hints for an untitled file in VS Code to disk

compiler/src/main.rs
```diff
@@ -1,6 +1,7 @@
 #![feature(async_closure)]
 #![feature(box_patterns)]
 #![feature(label_break_value)]
+#![feature(let_chains)]
```
JonasWanke: These are stabilized in Rust 1.64.0-nightly (`rustup update`)

compiler/src/language_server/mod.rs
```diff
@@ -217,73 +218,103 @@ impl LanguageServer for CandyLanguageServer {
     }
 
     async fn did_open(&self, params: DidOpenTextDocumentParams) {
-        let input = params.text_document.uri.into();
+        let project_directory = self.project_directory.read().await.clone().unwrap();
+        let module = Module::from_package_root_and_url(
+            project_directory,
+            params.text_document.uri,
+            ModuleKind::Code,
+        );
```
JonasWanke: Can this be moved into a utility function to avoid repetition?

compiler/src/language_server/utils.rs
```diff
@@ -49,62 +49,60 @@ impl HirError {
...
+            Package::User(_) | Package::External(_) => Url::from_file_path(
+                module
+                    .to_possible_paths()
+                    .unwrap()
+                    .into_iter()
+                    .find(|path| path.exists())
+                    .unwrap(),
+            )
+            .unwrap(),
+            Package::Anonymous { .. } => Url::parse(&format!("untitled")).unwrap(),
```
JonasWanke: This looses the file ID

JonasWanke (changes requested):

JonasWanke (approved):

PR #238, opened 2022-08-07
Remove Values and use appropriate maps

MarcelGarus:
This PR applies some optimizations to the VM.
Most notably, the following changed:

- Previously, each struct access such as `core.int` caused the entire core struct to be exported three times (for `TraceDebugValueEvaluated` of `core`, for `TraceCallStarted` of calling the struct access builtin, once by the builtin code itself). Because functions are just values in Candy, the concept of having "self-contained values" in the Rust sense leads to duplicate subtrees and is fundamentally flawed. That's why instead of using `Heap`-managed objects inside VMs and values for everything else, _all_ code now uses `Heap`s.
- We're more mindful about using `im::HashMap` only when necessary. In particular, the `Object::Struct` now uses our own struct implementation, which is mathematically less efficient but should serve us now.

Before these changes, the Core library couldn't run on my 32 GiB machine because it ran out of memory. It now runs in about one second, which is not great but at least a baseline for more performance optimizations. Ignoring the initial compilation of modules, an evaluation of `use "..Core"` with pre-compiled files in Salsa takes about 129 ms (using the very accurate benchmark also included in this PR 😁).


compiler/src/vm/tracer.rs
```diff
@@ -72,12 +72,13 @@ impl Tracer {
...
-    pub fn dump_stack_trace(&self, db: &Database) {
-        for line in self.format_stack_trace(db).lines() {
+    pub fn dump_stack_trace(&self, db: &Database, heap: &Heap) {
+        for line in self.format_stack_trace(db, heap).lines() {
             log::error!("{}", line);
         }
     }
-    pub fn format_stack_trace(&self, db: &Database) -> String {
+    pub fn format_stack_trace(&self, db: &Database, heap: &Heap) -> String {
+        // TODO: Format values properly.
```
JonasWanke: Looks like this is already done?

compiler/src/vm/tracer.rs
```diff
@@ -137,25 +141,35 @@ impl Tracer {
             .join("\n")
     }
 
-    pub fn dump_call_tree(&self) -> String {
+    pub fn format_call_tree(&self, heap: &Heap) -> String {
+        // TODO: Format values properly.
```
JonasWanke: Looks like this is already done?

compiler/src/vm/heap/pointer.rs
```diff
@@ -0,0 +1,21 @@
...
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
+pub struct Pointer(usize);
+
+impl Pointer {
+    pub fn null() -> Self {
+        Self(0)
+    }
+    pub fn from_raw(raw: usize) -> Self {
+        Self(raw)
+    }
```
JonasWanke, line 12: Maybe implement `From` instead?

compiler/src/vm/heap/pointer.rs
```diff
@@ -0,0 +1,21 @@
...
+    pub fn raw(&self) -> usize {
+        self.0
+    }
+}
+impl Display for Pointer {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
```
JonasWanke: Maybe format this as hexadecimal?

compiler/src/vm/builtin_functions.rs
```diff
@@ -1,373 +1,443 @@
...
-        destructure!(args, [Value::Closure(Closure { num_args, .. })], {
-            Ok((*num_args).into())
+    fn function_run(&mut self, args: &[Pointer]) -> BuiltinResult {
+        unpack_and_later_drop!(self, args, (closure: Closure), {
+            closure.should_take_no_arguments()?;
+            log::debug!(
+                "`functionRun` executing the closure: {}",
+                closure.address.format(self),
+            );
+            self.dup(closure.address);
```
JonasWanke: Is this only because `unpack_and_later_drop` would drop it otherwise? If so, I'd recommend to create a second macro that only only unpacks without dropping

compiler/src/vm/builtin_functions.rs
```diff
@@ -1,373 +1,443 @@
...
             }
         )
     }
 
-    fn int_add(&mut self, args: Vec<Value>) -> Result<Value, String> {
-        destructure!(args, [Value::Int(summand_a), Value::Int(summand_b)], {
-            Ok((summand_a + summand_b).into())
+    fn int_add(&mut self, args: &[Pointer]) -> BuiltinResult {
+        unpack_and_later_drop!(self, args, (a: Int, b: Int), {
+            Return(self.create_int(a.value.clone() + b.value.clone()))
```
JonasWanke: Do these really need to be cloned here and below? `Add` is implemented for references

compiler/src/vm/builtin_functions.rs
```diff
@@ -1,373 +1,443 @@
...
+            };
+            Return(self.create_symbol(symbol.to_string()))
         })
     }
 }
+
+trait ClosureShouldHaveNoArguments {
+    fn should_take_no_arguments(&self) -> Result<(), String>;
+}
+impl ClosureShouldHaveNoArguments for Closure {
```
JonasWanke: Do we need this trait if `Closure` is from the same crate?

compiler/src/vm/use_provider.rs
```diff
@@ -78,9 +76,9 @@ struct UsePath {
...
     const PARENT_NAVIGATION_CHAR: char = '.';
 
-    fn parse(path: Value) -> Result<Self, String> {
-        let path = match path {
-            Value::Text(path) => path,
+    fn parse(heap: &Heap, path: Pointer) -> Result<Self, String> {
+        let path = match &heap.get(path).data {
+            Data::Text(path) => path.value.clone(),
             _ => return Err("the path has to be a text".to_string()),
         };
```
JonasWanke: Can this use `try_into()`?

compiler/src/vm/heap/object.rs
```diff
@@ -0,0 +1,257 @@
...
+
+#[derive(Clone)]
+pub struct Text {
+    pub value: String,
+}
+
+#[derive(Clone)]
+pub struct Symbol {
+    // TODO: For better memory efficiency, store the symbol as a null-terminated
+    // string.
```
JonasWanke: Because it's a subset of ASCII, we could store it even more efficiently (5 bits for the first and 6 bits for the following characters are sufficient), and we should probably intern them

compiler/src/vm/heap/object.rs
```diff
@@ -0,0 +1,257 @@
...
+            if existing_key.equals(heap, key) {
+                *existing_value = value;
+                return;
+            }
+        }
+        let entry = (hash, key, value);
+        if let Some(index) = self
+            .fields
+            .iter()
+            .position(|(the_hash, _, _)| *the_hash > hash)
```
JonasWanke: If the fields are sorted, we could (should?) use binary search. Also, a single traversal is sufficient

compiler/src/vm/heap/object.rs
```diff
@@ -0,0 +1,257 @@
...
+                    return false;
+                }
+            }
+        }
+        true
+    }
+}
+
+impl Closure {
+    pub fn of_lir(module: Module, lir: Lir) -> Self {
```
JonasWanke:
```suggestion
    pub fn of_module_lir(module: Module, lir: Lir) -> Self {
```

compiler/src/vm/heap/object.rs
```diff
@@ -0,0 +1,257 @@
...
+    fn hash_with_state<H: Hasher>(&self, heap: &Heap, state: &mut H) {
+        match self {
+            Data::Int(int) => int.value.hash(state),
+            Data::Text(text) => text.value.hash(state),
+            Data::Symbol(symbol) => symbol.value.hash(state),
+            Data::Struct(struct_) => {
+                let mut s = 0;
+                for (key, value) in struct_.iter() {
+                    s ^= key.hash(heap);
+                    s ^= value.hash(heap);
```
JonasWanke, line 170: Shouldn't this be something like `key.hash_with_state(heap, state)` instead of manually XORing the hashes?

compiler/src/vm/heap/object.rs
```diff
@@ -0,0 +1,257 @@
...
+                "[{}]",
+                struct_
+                    .fields
+                    .iter()
+                    .map(|(_, key, value)| (key.format(heap), value.format(heap)))
+                    .sorted_by(|(key_a, _), (key_b, _)| key_a.cmp(key_b))
+                    .map(|(key, value)| format!("{}: {}", key, value))
+                    .join(", ")
+            ),
+            Data::Closure(_) => "{...}".to_string(),
```
JonasWanke:
```suggestion
            Data::Closure(_) => "{…}".to_string(),
```

compiler/src/vm/heap/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+#[derive(Clone)]
+pub struct Heap {
+    objects: HashMap<Pointer, Object>,
+    next_address: Pointer,
+}
+
+impl std::fmt::Debug for Heap {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut objects = self.objects.clone().into_iter().collect_vec();
+        objects.sort_by_key(|(address, _)| address.raw());
```
JonasWanke: Do we really have to clone the whole heap or could we, e.g., collect and sort the keys and then access values in the `for` loop below?

compiler/src/vm/heap/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub fn dup(&mut self, address: Pointer) {
+        self.get_mut(address).reference_count += 1;
+        log::trace!(
+            "RefCount of {address} increased to {}. Value: {}",
+            self.get(address).reference_count,
+            address.format(self),
+        );
+    }
+    pub fn drop(&mut self, address: Pointer) {
+        let formatted_value = address.format(self);
```
JonasWanke, line 68: Maybe move this into the `log::trace!(…)` to make it more explicit that it's only for logging? And I think you could also call `object.format(self)` instead to remove one lookup

compiler/src/vm/heap/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub fn create(&mut self, object: Data) -> Pointer {
+        let address = self.next_address;
+        self.objects.insert(
+            address,
+            Object {
+                reference_count: 1,
+                data: object,
+            },
+        );
+        log::trace!("Created object {} at {address}.", address.format(self));
```
JonasWanke: You could locally store the created `Object` and call it's `format(…)` directly to avoid the lookup

compiler/src/vm/heap/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    ) -> Vec<Pointer> {
+        let mut objects_to_refcounts = HashMap::new();
+        for address in addresses {
+            self.gather_objects_to_clone(&mut objects_to_refcounts, *address);
+        }
+        let num_objects = objects_to_refcounts.len();
+
+        let address_map: HashMap<Pointer, Pointer> = objects_to_refcounts
+            .keys()
+            .cloned()
```
JonasWanke:
```suggestion
            .copied()
```

compiler/src/vm/heap/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+    pub fn create_nothing(&mut self) -> Pointer {
+        self.create_symbol("Nothing".to_string())
+    }
+    pub fn create_list(&mut self, items: Vec<Pointer>) -> Pointer {
+        let mut fields = vec![];
+        for (index, item) in items.into_iter().enumerate() {
+            fields.push((self.create_int(index.into()), item));
+        }
+        self.create_struct(fields.into_iter().collect())
```
JonasWanke: We could insert them into a `HashMap` directly

compiler/src/vm/heap/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+        let fields = vec![
+            (
+                self.create_symbol("Type".to_string()),
+                self.create_symbol(type_),
+            ),
+            (self.create_symbol("Value".to_string()), value),
+        ]
+        .into_iter()
+        .collect();
+        self.create_struct(fields)
```
JonasWanke:
```suggestion
        let fields = HashMap::from([
            (
                self.create_symbol("Type".to_string()),
                self.create_symbol(type_),
            ),
            (self.create_symbol("Value".to_string()), value),
        ]);
        self.create_struct(fields)
```

packages/Benchmark.candy
```diff
@@ -0,0 +1,100 @@
+core = use "..Core"
```
JonasWanke: Is there more to this benchmark, e.g., measuring?

JonasWanke (changes requested):

compiler/src/vm/builtin_functions.rs
MarcelGarus: Sadly, we do. Otherwise, the compiler complains because addition from Rust's [`Add` trait](https://doc.rust-lang.org/std/ops/trait.Add.html) has the signature `fn add(self, rhs: Rhs)`.

compiler/src/vm/heap/mod.rs
MarcelGarus: A problem is that the `object` is mutable (and because the object is borrowed from `self`, it thus represents a mutable reference to `self`), so we can't at the same time borrow `self` immutably for formatting the object. That's why it's a bit clunky here.

compiler/src/vm/heap/mod.rs
MarcelGarus: Just to be clear: You propose formatting the object before moving it into the heap and then only outputting the string here? Or putting a clone of the object into the heap and then formatting the original in the trace?

compiler/src/vm/heap/object.rs
MarcelGarus: I tried to make sure that the ordering of the fields doesn't matter, because multiple structs considered equal may not return items in the same order (two keys with the same hash may be inserted in different orders)

compiler/src/vm/heap/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+            .unwrap_or_else(|| panic!("Couldn't get object {address}."))
+    }
+
+    pub fn dup(&mut self, address: Pointer) {
+        self.get_mut(address).reference_count += 1;
+        log::trace!(
+            "RefCount of {address} increased to {}. Value: {}",
+            self.get(address).reference_count,
+            address.format(self),
+        );
```
JonasWanke: We can still reduce one lookup here:

```suggestion
        let object = self.get(address);
        trace!(
            "RefCount of {address} increased to {}. Value: {}",
            object.reference_count,
            object.format(self),
        );
```

compiler/src/vm/heap/mod.rs
JonasWanke: Either formatting it before it gets moved into the heap, or we just swap the order:

```rust
        let object = Object {
            reference_count: 1,
            data: object,
        };
        trace!("Creating object {} at {address}.", object.format(self));
        self.objects.insert(address, object);
        self.next_address = Pointer::from_raw(self.next_address.raw() + 1);
```

compiler/src/vm/heap/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            Object {
+                reference_count: 1,
+                data: object,
+            },
+        );
+        log::trace!("Created object {} at {address}.", address.format(self));
+        self.next_address = Pointer::from_raw(self.next_address.raw() + 1);
+        address
+    }
+    pub fn free(&mut self, address: Pointer) {
```
JonasWanke: Does this need to be `pub`?

compiler/src/vm/heap/object.rs
```diff
@@ -0,0 +1,238 @@
...
+#[derive(Clone)]
+pub struct Symbol {
+    // TODO: Choose a more efficient representation.
+    pub value: String,
+}
+
+#[derive(Default, Clone)]
+pub struct Struct {
+    // TODO: Choose a more efficient representation.
+    pub buckets: HashMap<u64, Vec<(Pointer, Pointer)>>, // hash -> list of keys and values
```
JonasWanke: I think I'd rather revert this to a single `Vec<u64, Pointer, Pointer>`. I'm assuming that our structs usually contain few keys so a `HashMap` is probably overkill, especially if we sort the entries. Also, this `HashMap` representation with `Vec`s adds 24 bytes and another level of indirection to each entry with distinct hashes.

With a single, sorted `Vec`, we could reduce the previous two linear passes to one binary search plus a linear pass within hash collisions by sorting the struct by hash and then the raw key's pointer:

```rust
        let hash = key.hash(heap);
        let first_index_for_hash = self.fields
            .binary_search_by_key(
                (&hash, Pointer::null()),
                |(hash, key, _)| (hash, key),
            )
            .into_ok_or_error();
            // needs `#![feature(result_into_ok_or_err)]`
            // https://github.com/rust-lang/rust/issues/82223
        let mut index = first_index_for_hash;
        for (existing_index, (existing_hash, existing_key, existing_value)) in &mut self.fields[first_index_for_hash..].iter_mut().enumerate() {
            if hash != *existing_hash {
                index = existing_index;
                break;
            }
            if existing_key.equals(heap, key) {
                *existing_value = value;
                return;
            }
            if existing_key.raw < key.raw {
                index = existing_index;
            }
        }
        self.fields.insert(existing_index, (hash, key, value));
```

compiler/src/vm/heap/object.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub fn iter(&self) -> impl Iterator<Item = (Pointer, Pointer)> {
+        self.buckets.clone().into_values().flatten()
+    }
+    fn equals(&self, heap: &Heap, other: &Struct) -> bool {
+        if self.len() != other.len() {
+            return false;
+        }
+        for (key_a, value_a) in self.iter() {
+            for (key_b, value_b) in other.iter() {
+                if key_a.equals(heap, key_b) && !value_a.equals(heap, value_b) {
```
JonasWanke: Different keys also mean that the structs are not equal

```suggestion
                if !key_a.equals(heap, key_b) || !value_a.equals(heap, value_b) {
```

compiler/src/vm/builtin_functions.rs
```diff
@@ -78,6 +78,27 @@ impl From<SuccessfulBehavior> for BuiltinResult {
...
+                );
+            };
+            let ( $( $arg, )+ ): ( $( UnpackedData<$type>, )+ ) = ( $(
+                UnpackedData {
+                    address: $arg,
+                    data: $heap.get($arg).data.clone().try_into()?,
+                },
+            )+ );
+
+            $body.into()
```
JonasWanke, line 98: Can we reuse this in `unpack_and_later_drop!(…)`?

JonasWanke (changes requested):

compiler/src/vm/builtin_functions.rs
MarcelGarus: For some reason, this doesn't work.
In particular, this causes variables in arguments to not be resolved.

image

compiler/src/vm/heap/object.rs
MarcelGarus: Hmm. Our values are not interned, so we can't simply search by pointers – if I'm looking for whatever is at key `123`, we want to consider that key equal to another `123` with a different address. So, we'd have to use `binary_search_by` and do the comparison ourselves in a closure. Except, we can't compare all Candy values. In particular, closures have no well-defined ordering and I don't think they should.

We could of course only binary search for the hash. Then, once we found a matching hash, we look in both directions of the `Vec` (assuring the hash is still the same) and compare our value to the others that exist there. Semantically, this is similar to the current approach (first looking for a bucket, then comparing to elements in it), performance-wise it's probably better for read-heavy structs (which all Candy structs are because we don't support mutation). Do you think it's worth implementing this now or should we leave that for a later time? I can imagine it'll be a lot more code than the current 6-line function.

JonasWanke: Have a look at the code I posted above. It implements what you describe in your second paragraph: We do a binary search to find the first matching hash in the struct and then traverse linearly to find the matching key. The key ordering is only based on the pointer values, not the actual key values. 

It's not that much code, though I haven't tested it and the getter also has to be modified

MarcelGarus: If there are multiple matches, then `binary_search_by_key` returns any one of them, not the first one. But I guess I can implement it this way

MarcelGarus: Ah nevermind, I thought that was the code for retrieving an item

JonasWanke: Yes, that's why I'm searching for the hash in combination with a null pointer: That way, we always find the left-most element with a matching hash (if it exists)

compiler/src/vm/heap/object.rs
```diff
@@ -73,38 +72,57 @@ impl Struct {
...
+            .iter()
+            .enumerate()
+            .take_while(|(_, (existing_hash, _, _))| *existing_hash == hash)
+            .map(|(i, (_, key, _))| (index_of_hash + i, key));
+        let going_left = self.fields[0..index_of_hash]
+            .iter()
+            .rev()
+            .enumerate()
+            .take_while(|(_, (existing_hash, _, _))| *existing_hash == hash)
+            .map(|(i, (_, key, _))| (index_of_hash - i - 1, key));
```
JonasWanke: You can simplify this by searching for `(hash, Pointer::null)`: This always finds the left-most index for the given hash. This way, you only have to search in one direction from there.

compiler/src/vm/heap/object.rs
```diff
@@ -73,38 +72,57 @@ impl Struct {
...
     pub fn get(&self, heap: &Heap, key: Pointer) -> Option<Pointer> {
-        let bucket = self.buckets.get(&key.hash(heap))?;
-        for (existing_key, existing_value) in bucket {
-            if existing_key.equals(heap, key) {
-                return Some(*existing_value);
-            }
-        }
-        None
+        let index = self.index_of_key(heap, key).ok()?;
+        Some(self.fields[index].1)
```
JonasWanke: `.1` is the key, not the value

JonasWanke (changes requested):

compiler/src/vm/heap/object.rs
MarcelGarus: Ohhhh I didn't catch that. Makes sense (although I found an even more readable version)

MarcelGarus:
I also updated the benchmark to something compute-heavy that's not just a bunch of code cloning. The benchmark now calculates a Fibonacci number in under a second (789 ms).

compiler/src/vm/heap/object.rs
```diff
@@ -76,36 +76,32 @@ impl Struct {
...
-        let going_left = self.fields[0..index_of_hash]
-            .iter()
-            .rev()
-            .enumerate()
-            .take_while(|(_, (existing_hash, _, _))| *existing_hash == hash)
-            .map(|(i, (_, key, _))| (index_of_hash - i - 1, key));
+            .map(|(i, (_, key, _))| (index_of_first_hash_occurrence + i, key));
 
-        for (index, existing_key) in going_left.chain(going_right) {
+        for (index, existing_key) in fields_with_same_hash {
```
JonasWanke, line 87: We could even perform a binary search within entries with the same hash in case the key has the same pointer, but I don't think that would be worth it. Also, we'd have to sort the entries by key pointers as well

compiler/src/vm/heap/object.rs
```diff
@@ -76,36 +76,32 @@ impl Struct {
...
             }
         }
-        Err(index_of_hash)
+        Err(index_of_first_hash_occurrence)
     }
     fn insert(&mut self, heap: &Heap, key: Pointer, value: Pointer) {
-        let index = self.index_of_key(heap, key);
-        self.fields
-            .insert(index.into_ok_or_err(), (key.hash(heap), key, value));
+        let field = (key.hash(heap), key, value);
```
JonasWanke: We hash the key here and in `index_of_key`, maybe we could pass the hash to the lookup?

JonasWanke (approved):

PR #239, opened 2022-08-07
Better logging

MarcelGarus:
Depends on #238

Standard logging can get confusing when used with async/await because a series of events doesn't have to be connected. This PR adds structured logging using the [tracing crate](https://crates.io/crates/tracing), which is a de-facto standard for structured logging and is also supported by salsa.

Like with usual logging, you can use the `info!` etc. macros to emit events. But it also supports _spans_, which log ongoing operations. All events logged within a span then contain the span for context. This is not only useful for logs of multiple parallel language server commands, but also for debugging our parser:

image

Or for measuring the time of the compilation stages (the `enter` and `exit` events mean the corresponding future actually executed code vs. waited for I/O or other stuff):

image

This is a bit verbose though, so I disabled it for now.

Theoretically, we could also log values directly (instead of formatting them to a string) and then dynamically filter the logs based on the values, but that's probably a bit overkill for now.

---

Also, even if we filter logs to a debug level, the format strings for tracing are still created, which can be pretty expensive. In particular, the formatting of values in the heap is unnecessary (values may be large structs) if the logs are filtered out anyways.
That's why I also added `TRACE` constants to the VM and heap that you can manually set to `true` if you want to generate traces. That's not super elegant, but more elegant than commenting out the code (which is what was done before in the VM).

Using this, the `use "..Core"` benchmark drops from 111 ms to 103 ms.

(See comment below for next steps.)


MarcelGarus:
The remaining elephant in the room is the copying of instructions – the flamegraph is dominated by clones and drops of instructions (or closures, which contain lots of instructions):

image

Apart from that, the VM is now reasonably fast – at least, there are no obvious parts where we have too much algorithmic complexity (such as the `Value` before) or do unnecessary work (such as with the logging).
This is supported by numbers: Our VM currently executes `use "..Core"` in ~100 ms, excluding the compilation of files (at least kind of, I assume that compilation doesn't play a huge part when averaging over hundred invocations, 99 of which don't need to compile anything new and are served by salsa). Running `use "..Core"` takes ~80000 instructions, so our VM runs about 800000 instructions per second.

Regarding performance, the next sensible step is probably to do some basic optimizations on the intermediary representations (aka reduce the number of instructions from 80000 to a lower number by doing some compile-time evaluation including caching, eliminating common pure subtrees, inlining functions, etc.) Next to having to execute fewer instructions, this also means that fewer instructions need to be copied (aka reducing the biggest factor in the flamegraph).
The optimization with the potentially biggest payoff is using compile-time evaluation to simplify code. However, I suggest we delay further optimizing the VM until after we implemented fibers and channels (i.e. we model side-effects explicitly) because handling side-effect-ful code is something that affects our optimization strategies.

compiler/src/fuzzer/mod.rs
```diff
@@ -6,48 +6,49 @@ pub use self::fuzzer::{Fuzzer, Status};
...
-                    arguments.iter().map(|it| format!("{}", it)).join(" "),
+                    arguments
+                        .iter()
+                        .map(|argument| argument.format(heap))
+                        .join(" "),
                 );
-                log::error!("This was the stack trace:");
-                tracer.dump_stack_trace(db);
+                error!("This was the stack trace:");
+                tracer.dump_stack_trace(db, heap);
```
JonasWanke: What do you think about lowering the severity to `info`? It's not an actual error of our compiler, but just an error in the user's code that got discovered

compiler/src/main.rs
```diff
@@ -215,50 +234,55 @@ async fn fuzz(options: CandyFuzzOptions) {
...
+        .with_filter(filter::filter_fn(level_for("candy::vm", Level::DEBUG)))
+        .with_filter(filter::filter_fn(level_for(
+            "candy::vm::heap",
+            Level::DEBUG,
+        )));
+    tracing_subscriber::registry().with(console_log).init();
+}
+fn level_for(module: &'static str, level: Level) -> impl Fn(&Metadata) -> bool {
+    move |metadata| {
+        if metadata.module_path().unwrap_or("").starts_with(module) {
```
JonasWanke:
```suggestion
        if metadata.module_path().unwrap_or_default().starts_with(module) {
```

JonasWanke (changes requested):

JonasWanke (approved):

JonasWanke:
(But Clippy complains about an unused import)

PR #241, opened 2022-09-11
Channels, fibers, oh my!

MarcelGarus:
This PR adds channels, fibers, `try`, and environments.

Here are all the new features:

- `core.try { ... }` catches panics of the closure and instead returns a result, either `[Ok, returnValue]` or `[Error, panicReason]`.

- `core.parallel { nursery -> ... }` spawns a new parallel scope, in which multiple fibers may run concurrently. Which brings us to the next feature:

- `core.async nursery { ... }` spawns a new fiber and returns a future (which is just a receive port). `core.await someFuture` waits for its result. In a way, `core.async` and `core.await` are the opposite of each other (`core.await (core.async nursery foo) == foo`).
  Note that panics still propagate upwards – a panicking fiber that was spawned on a nursery also causes the surrounding `core.parallel` to panic. All other fibers spawned on that nursery get canceled (they are not further executed).

- The `candy run` command now looks for an exported `main` function and calls it with an environment.

Here's everything in action:

```candy
main environment =
  print message = core.channel.send environment.stdout message

  core.parallel { nursery ->
    core.async nursery { print "Hello, world!" }
    core.async nursery { print "Hallo, Welt!" }
    core.async nursery { print "Hola, mundo!" }

    four = core.async nursery { 4 }
    core.await four
  }

# Executing this prints that the main function returned `4`.
```

**Implementation**

Although we already spoke about my initial idea in person, I completely changed my approach since then.

Initially, I planned on implementing a fiber tree using an actual tree of Rust values. However, I didn't manage to implement that in a reasonable amount of time (I tried for ~20 hours total). Instead, I ended up with a very simple approach, inspired by the implementation of the BeamVM. In this approach, there's not an actual tree of structs. Rather, a VM maintains a list of fibers as well as a list of channels.

The main advantage of this new approach is the simplicity of the implementation. In particular, there's less accidental complexity that arises from organizing channels and fibers in a tree. For example, no migrating of channels between different subtrees is necessary because all channels are global to the VM anyway. This especially reduces the complexity in Rust, where ownership can make things a _lot_ less complicated.

Looking forward, this architecture also enables straightforward parallelization. We don't have to lock subtrees of fibers, but instead, we can have a pool of worker threads that try to work on the fibers, for example, by choosing a random fiber from the list, and locking it (if no one else is working on it already).
A more advanced approach (also implemented by the BeamVM) seems to be to have fibers assigned to a particular worker thread (similar to the affinity of OS threads) because this improves the cache hit rate (L1 caches are local per core). This is often combined with work stealing. I'm sure we can implement something similar for our VM.

**Performance**

Entering parallel scopes and creating new fibers is often veeery slow. The reason for that is that if `core` is used inside a fiber, then the whole `core` struct is copied to another fiber. This is not an exponential blowup of runtime like the value-issue we had previously, but fibers are definitely not cheap.

In the future, several optimizations can improve the runtime:

- Compile-time evaluation and inlining can reduce the amount of code cloned to fibers. For example, a `core.int.add` may be reduced to `✨.intAdd`, so that only `✨` needs to be copied to the fiber.
- Constant heaps can allow code to be shared. Compile-time evaluated values don't need to be manually constructed using instructions like `createInt` and `createStruct`. Rather, we could have a separate heap area that contains constant objects. Several other compilers also use this technique (e.g. Lua, Wren). The objects in that heap could have a special header that indicates they are not reference-counted, so `dup` and `drop` instructions on those objects are a no-op.
- Share heaps? Sometimes? It may sometimes make sense to share heaps between multiple fibers. This makes every reference count operation a lot more expensive (they need to be atomic, synchronized between CPU caches, etc.), but if we access very few, random-like bits of a big data structure, this may make sense. For the common case of fibers that work on mostly independent tasks, this handling would only slow them down though. So it depends a lot on the access pattern of fibers.


MarcelGarus:
Okay, I think my todos are done. The only unpolished part is the tracer. It's not worse than before, but it's far from finished: For now, the tracer returned by the VM's `tear_down` method only returns the tracer of the outermost fiber, and all the inner ones are ignored/dropped.

The main reason is that its whole architecture relied on being owned by a fiber (for example, it only stores pointers because it doesn't have its own heap).

I also noticed that the way we use the tracer for determining whose fault a panic is isn't quite correct all the time. In particular, we need to handle cases where `needs` outlive their call sites:

```candy
tripleIntGenerator intGenerator =
  safeIntGenerator = {
    result = call intGenerator
    needs (isInt result) "tripleIntGenerator only accepts closures that always return ints"
    result
  }
  { multiply 3 (safeIntGenerator call) }

intGenerator = { "Bad return value" }
foo = tripleIntGenerator intGenerator
bar = foo call
```

Here, the `tripleIntGenerator` takes an int-producing closure and returns another closure that always produces trice as much. Because the `needs` is defined in an internal closure, if it fails that's the fault of the caller of `tripleIntGenerator`, _not_ the caller of the closure. As a consequence, the assignment of `bar` panics. However, we currently can't determine that the fault is in the line where we create the `foo` (because that's the part of the code that calls `tripleIntGenerator` with a non-int closure). Ideally, we would highlight that line in the editor.

I have ideas on how to solve this but I'd rather do that in a separate PR because it requires more work, and this PR is already quite large. In the next PR, I'll separate the tracer into two parts:

- A fault analyzing component owned by the fibers. Essentially, fibers would have an additional fault stack (next to the data, call, and import stacks) which contains places of code that can be at fault (I assume the HIR IDs of calls). The top-most item is the one that's at fault if a need panics. Closures would also capture the top value of that stack and when executed, put that on the stack again and remove it once they're done. This would (finally) allow accurately and explicitly finding out who's at fault when code crashes.

- The tracing part, which would be moved outside the fibers and only borrowed to running fibers. Semantically, that better fits the perspective that fibers report to some service as a side effect, simplifying fibers themselves to state machines only containing essential complexity. Because the VM would own the tracer and provide it to fibers, the VM can also report other events such as fibers being entered and exited during scheduling or channel operations.

README.md
```diff
@@ -117,22 +117,27 @@ Major milestones:
 
 ## Short-term TODOs
 
-- implement fibers, channels, and nurseries
-- remove builtinPrint
+- casing of module names
```
JonasWanke: Already merged :)

README.md
```diff
@@ -117,22 +117,27 @@ Major milestones:
...
-- support recursion
+- remove builtinPrint
 - tail call optimization
-- new name?
 - parse function declaration with doc comment but no code
+- tracing visualization
+- distinguish packages from normal modules
 - complain about comment lines with too much indentation
 - develop guidelines about how to format reasons
+- module files: `.candy` vs `_.candy`?
```
JonasWanke: Already merged :)

compiler/src/vm/channel.rs
```diff
@@ -0,0 +1,64 @@
+use super::{Heap, Pointer};
+use std::{collections::VecDeque, fmt};
+
+/// A conveyer belt or pipe that flows between send and receive ports in the
+/// program. Using send ports, you can put packets into a channel. Using receive
+/// ports, you can get packets out again.
+///
+/// Channels always have a maximum capacity of packets that they can hold
+/// simultaneously – you can set it to something large, but having no capacity
```
JonasWanke:
```suggestion
/// simultaneously – you can set it to something large, but having no limit
```

compiler/src/vm/channel.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.packets.is_empty()
+    }
+    pub fn is_full(&self) -> bool {
+        self.packets.len() == self.capacity
+    }
+
+    pub fn send(&mut self, packet: Packet) {
+        if self.is_full() {
+            panic!("Tried to send on channel that is full.");
```
JonasWanke:
```suggestion
            panic!("Tried to send on a channel that is full.");
```

compiler/src/vm/channel.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+        self.packets.push_back(packet);
+    }
+
+    pub fn receive(&mut self) -> Packet {
+        self.packets
+            .pop_front()
+            .expect("Tried to receive from channel that is empty.")
```
JonasWanke:
```suggestion
            .expect("Tried to receive from a channel that is empty.")
```

compiler/src/vm/fiber.rs
```diff
@@ -1,36 +1,48 @@
...
     pub status: Status,
     next_instruction: InstructionPointer,
-    pub heap: Heap,
     pub data_stack: Vec<Pointer>,
     pub call_stack: Vec<InstructionPointer>,
     pub import_stack: Vec<Module>,
+    pub heap: Heap,
+
+    // Debug stuff. This is not essential to a correct working of the fiber, but
+    // enables advanced functionality like stack traces or finding out who's
```
```suggestion
    // enables advanced functionality like stack traces or finding out whose
```

compiler/src/vm/fiber.rs
```diff
@@ -127,18 +136,77 @@ impl Vm {
         self.status.clone()
     }
 
+    // If the status of this fiber is something else than `Status::Running`
+    // after running, then the VM that manages this fiber is expected to perform
+    // some action and to then call the corresponding `complete_*` method before
+    // calling `run` again.
+
+    pub fn complete_channel_create(&mut self, channel: ChannelId) {
```
JonasWanke, line 145: Can we add asserts here and for the functions below that the call matches the current state?

compiler/src/vm/fiber.rs
```diff
@@ -127,18 +136,77 @@ impl Vm {
...
+        self.status = Status::Running;
+    }
+    pub fn complete_receive(&mut self, packet: Packet) {
+        let address = packet
+            .heap
+            .clone_single_to_other_heap(&mut self.heap, packet.value);
+        self.data_stack.push(address);
+        self.status = Status::Running;
+    }
+    pub fn complete_parallel_scope(&mut self, result: Result<(), String>) {
```
JonasWanke: Does anything speak against letting the parallel scope return a value directly?

compiler/src/vm/fiber.rs
```diff
@@ -127,18 +136,77 @@ impl Vm {
...
+    }
+    pub fn complete_try(&mut self, result: Result<Packet, String>) {
+        let result = match result {
+            Ok(Packet {
+                heap,
+                value: return_value,
+            }) => {
+                let ok = self.heap.create_symbol("Ok".to_string());
+                let return_value = heap.clone_single_to_other_heap(&mut self.heap, return_value);
+                self.heap.create_list(&[ok, return_value])
```
JonasWanke: We have `heap.create_result(…)`

compiler/src/vm/fiber.rs
```diff
@@ -127,18 +136,77 @@ impl Vm {
...
+            }
+            Err(panic_reason) => {
+                let err = self.heap.create_symbol("Err".to_string());
+                let reason = self.heap.create_text(panic_reason);
+                self.heap.create_list(&[err, reason])
```
JonasWanke: We have `heap.create_result(…)`

compiler/src/vm/fiber.rs
```diff
@@ -127,18 +136,77 @@ impl Vm {
...
+        };
+        self.data_stack.push(result);
+        self.status = Status::Running;
+    }
+
     fn get_from_data_stack(&self, offset: usize) -> Pointer {
         self.data_stack[self.data_stack.len() - 1 - offset as usize]
     }
+    pub fn panic(&mut self, reason: String) {
+        self.status = Status::Panicked { reason };
```
JonasWanke, line 205: This could also have an assert that the fiber was running before

compiler/src/vm/heap/mod.rs
```diff
@@ -214,16 +232,21 @@ impl Heap {
...
-            .enumerate()
-            .map(|(index, item)| (self.create_int(index.into()), item))
-            .collect();
-        self.create_struct(fields)
+    pub fn create_list(&mut self, items: &[Pointer]) -> Pointer {
+        let mut fields = vec![];
+        for (index, item) in items.iter().enumerate() {
+            fields.push((self.create_int(index.into()), *item));
+        }
+        self.create_struct(fields.into_iter().collect())
```
JonasWanke: What does this improve over the previous implementation? Specifically, since we now collect a list before converting that into a map

compiler/src/vm/heap/object.rs
```diff
@@ -151,6 +154,40 @@ impl Closure {
...
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct ChannelId(usize);
+
+impl From<usize> for ChannelId {
+    fn from(id: usize) -> Self {
+        Self(id)
+    }
+}
```
JonasWanke: I think `1.into()` could be too implicit and would rather go for an explicit `ChannelId::from(1)` without implementing the `From` trait

packages/Core/Concurrency.candy
```diff
@@ -0,0 +1,20 @@
+channel = use "..Channel"
```
JonasWanke: The `.candy` files have to be updated to the new naming scheme

packages/Core/Channel.candy
```diff
@@ -0,0 +1,18 @@
+isInt = (use "..Int").is
+equals = (use "..Equality").equals
```
JonasWanke: If we want to keep them ordered alphabetically

```suggestion
equals = (use "..Equality").equals
isInt = (use "..Int").is
```

packages/Core/Concurrency.candy
```diff
@@ -0,0 +1,20 @@
...
+function = use "..Function"
+structGet = (use "..Struct").getUnwrap
```
JonasWanke: To avoid confusion (references below have to be updated as well)

```suggestion
structGetUnwrap = (use "..Struct").getUnwrap
```

compiler/src/vm/fiber.rs
```diff
@@ -127,18 +136,77 @@ impl Vm {
...
+    // If the status of this fiber is something else than `Status::Running`
+    // after running, then the VM that manages this fiber is expected to perform
+    // some action and to then call the corresponding `complete_*` method before
+    // calling `run` again.
+
+    pub fn complete_channel_create(&mut self, channel: ChannelId) {
+        let send_port = self.heap.create_send_port(channel);
+        let receive_port = self.heap.create_receive_port(channel);
+        self.data_stack
+            .push(self.heap.create_list(&[send_port, receive_port]));
```
JonasWanke: Instead of a pair, a named tuple like `[Send: sendPort, Receive: receivePort]` (or maybe `[SendPort: sendPort, ReceivePort: receivePort]`  since they aren't lambdas but just ports) could avoid confusion

packages/Benchmark.candy
```diff
@@ -3,12 +3,36 @@
...
+#       fibRec fibRec (core.int.subtract n 1)
+#       fibRec fibRec (core.int.subtract n 2)
+#   }
+# fib n = fibRec fibRec n
+# 
+# twentyOne := fib 8
+
+channel = core.channel.create 3
+sendPort = core.struct.getUnwrap channel 0
+receivePort = core.struct.getUnwrap channel 1
```
JonasWanke: If `channel.create` is changed to return a map instead of a list, this can be simplified (and `core.async` as well)

```suggestion
channel = core.channel.create 3
sendPort = channel.send # or better `channel.sendPort` to avoid confusion?
receivePort = channel.receive
```

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+    /// The fiber of this tree entered a `core.parallel` scope so that it's now
+    /// paused and waits for the parallel scope to end. Instead of the main
+    /// former single fiber, the tree now runs the closure passed to
+    /// `core.parallel` as well as any other spawned children.
+    Parallel(Parallel),
+
+    Try(Try),
+}
+
+/// Single fibers are the leafs of the fiber tree.
```
JonasWanke:
```suggestion
/// Single fibers are the leaves of the fiber tree.
```

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+        self.fibers.insert(
+            root_fiber_id,
+            FiberTree::Single(Single {
+                fiber,
+                parent: None,
+            }),
+        );
+        self.root_fiber = Some(root_fiber_id);
+    }
+    pub fn set_up_for_running_closure(
```
JonasWanke, line 143: Why aren't these constructors?

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+    }
+
+    // pub fn fiber(&self) -> &Fiber {
+    //     // TODO: Remove before merging the PR
+    //     todo!()
+    // }
+    // pub fn cloned_tracer(&self) -> Tracer {
+    //     // TODO: Remove
+    //     self.fiber().tracer.clone()
+    // }
```
JonasWanke: TODOs

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn complete_receive(&mut self, performing_fiber: Option<FiberId>, packet: Packet) {
+        if let Some(fiber) = performing_fiber {
+            let tree = self.fibers.get_mut(&fiber).unwrap();
+            tree.as_single_mut().unwrap().fiber.complete_receive(packet);
+        }
+    }
```
JonasWanke: Why is the `performing_fiber` optional – can't the caller handle that?

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+                    id
+                };
+
+                self.fibers.replace(fiber_id, |tree| {
+                    let single = tree.into_single().unwrap();
+                    FiberTree::Parallel(Parallel {
+                        paused_fiber: single,
+                        children: [(first_child_id, ChildKind::InitialChild)]
+                            .into_iter()
+                            .collect(),
```
JonasWanke: There's a `hash_map!(…)` macro

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+                // directly, we communicate to the outside that no fiber
+                // references them anymore. The outside can then call
+                // `free_channel` when it doesn't intend to re-use the channel.
+                Channel::External => {
+                    self.push_external_operation(*channel, Operation::Drop);
+                }
+                // Nurseries are automatically removed when they are exited.
+                Channel::Nursery(_) => {}
+            }
+        }
```
JonasWanke: We should probably call this less often and optimize this in the future

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+    fn send_to_channel(
+        &mut self,
+        performing_fiber: Option<FiberId>,
+        channel_id: ChannelId,
+        packet: Packet,
+    ) {
+        let channel = match self.channels.get_mut(&channel_id) {
+            Some(channel) => channel,
+            None => {
+                // The channel was a nursery that died.
```
JonasWanke, line 474: How can this occur? If someone passes the nursery as a normal channel outside the parallel section? And are we sure, that the `performing_fiber` must be a single then?

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+                            .children
+                            .insert(child_id, ChildKind::SpawnedChild(return_channel));
+                    }
+                    None => self.finish_parallel(
+                        parent_id,
+                        Err("a nursery received an invalid message".to_string()),
+                    ),
+                }
+
+                InternalChannel::complete_send(&mut self.fibers, performing_fiber);
```
JonasWanke: Why is this call necessary?

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+            .get(return_channel_address)
+            .data
+            .clone()
+            .try_into()
+            .ok()?;
+
+        Some((heap, closure_address, return_channel.channel))
+    }
+
+    fn receive_from_channel(&mut self, performing_fiber: Option<FiberId>, channel: ChannelId) {
```
JonasWanke: This is only called from one place where the `performing_fiber` is set to `Some(fiber_id)`. Hence, `performing_fiber` can be simplified to a `FiberId`

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+    }
+
+    fn receive(
+        &mut self,
+        fibers: &mut HashMap<FiberId, FiberTree>,
+        performing_fiber: Option<FiberId>,
+    ) {
+        self.pending_receives.push_back(performing_fiber);
+        self.work_on_pending_operations(fibers);
+    }
```
JonasWanke: When would the `performing_fiber` be `None`?

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+    fn complete_receive(
+        fibers: &mut HashMap<FiberId, FiberTree>,
+        fiber: Option<FiberId>,
+        packet: Packet,
+    ) {
+        if let Some(fiber) = fiber {
+            let fiber = fibers.get_mut(&fiber).unwrap().as_single_mut().unwrap();
+            fiber.fiber.complete_receive(packet);
+        }
+    }
```
JonasWanke: I'd move the optional handling to the callers as well in case it's actually necessary

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+            Operation::Drop => write!(f, "dropping"),
+        }
+    }
+}
+
+impl From<usize> for FiberId {
+    fn from(id: usize) -> Self {
+        Self(id)
+    }
+}
```
JonasWanke: `From` might be too implicit

compiler/src/vm/mod.rs
```diff
@@ -1,8 +1,854 @@
...
+
+trait ReplaceHashMapValue<K, V> {
+    fn replace<F: FnOnce(V) -> V>(&mut self, key: K, replacer: F);
+}
+impl<K: Eq + Hash, V> ReplaceHashMapValue<K, V> for HashMap<K, V> {
+    fn replace<F: FnOnce(V) -> V>(&mut self, key: K, replacer: F) {
+        let value = self.remove(&key).unwrap();
+        let value = replacer(value);
+        self.insert(key, value);
```
JonasWanke, line 701: `self.entry(key).and_modify(…).or_insert_with(|| <panic>)` to avoid duplicate map traversion

compiler/src/language_server/hints/constant_evaluator.rs
```diff
@@ -75,46 +78,40 @@ impl ConstantEvaluator {
...
-                        Some(ast) => ast,
-                        None => continue,
-                    };
-                    if !matches!(ast.kind, AstKind::Assignment(_)) {
-                        continue;
-                    }
-                    (id.clone(), value)
+        let TearDownResult { heap, tracer, .. } = vm.clone().tear_down();
+        for entry in &tracer.events {
+            if let EventData::ValueEvaluated { id, value } = &entry.data {
```
JonasWanke: You can remove one indentation level with `let … else …`:

```suggestion
            let EventData::ValueEvaluated { id, value } = &entry.data else { continue; };
```

compiler/src/main.rs
```diff
@@ -200,30 +205,116 @@ fn run(options: CandyRunOptions) {
...
+    vm.set_up_for_running_module_closure(module_closure);
+    loop {
+        info!("Tree: {:#?}", vm);
+        match vm.status() {
+            Status::CanRun => {
+                debug!("VM still running.");
+                vm.run(&mut DbUseProvider { db: &db }, &mut RunForever);
+            }
+            Status::WaitingForOperations => {
+                todo!("VM can't proceed until some operations complete.");
```
JonasWanke: Can this actually occur here? This doesn't have access to the environment and hence can't create external operations, right?

compiler/src/main.rs
```diff
@@ -200,30 +205,116 @@ fn run(options: CandyRunOptions) {
...
-
-    let use_provider = DbUseProvider { db: &db };
-    let vm = Vm::new_for_running_module_closure(&use_provider, module_closure);
+    let mut vm = Vm::new();
+    vm.set_up_for_running_module_closure(module_closure);
+    loop {
+        info!("Tree: {:#?}", vm);
+        match vm.status() {
+            Status::CanRun => {
+                debug!("VM still running.");
```
JonasWanke:
```suggestion
                debug!("VM is still running.");
```

compiler/src/main.rs
```diff
@@ -200,30 +205,116 @@ fn run(options: CandyRunOptions) {
...
+    };
+
+    info!("Running main function.");
+    // TODO: Add environment stuff.
+    let mut vm = Vm::new();
+    let stdout = vm.create_channel();
+    let environment = {
+        let stdout_symbol = heap.create_symbol("Stdout".to_string());
+        let stdout_port = heap.create_send_port(stdout);
+        heap.create_struct([(stdout_symbol, stdout_port)].into_iter().collect())
```
JonasWanke: There's a `hash_map!(…)` macro

compiler/src/main.rs
```diff
@@ -200,30 +205,116 @@ fn run(options: CandyRunOptions) {
...
+    };
+    vm.set_up_for_running_closure(heap, main, &[environment]);
+    loop {
+        info!("Tree: {:#?}", vm);
+        match vm.status() {
+            Status::CanRun => {
+                debug!("VM still running.");
```
JonasWanke:
```suggestion
                debug!("VM is still running.");
```

compiler/src/main.rs
```diff
@@ -200,30 +205,116 @@ fn run(options: CandyRunOptions) {
...
+                todo!("VM can't proceed until some operations complete.");
+            }
+            _ => break,
+        }
+        let stdout_operations = vm
+            .external_operations
+            .get_mut(&stdout)
+            .unwrap()
+            .drain(..)
+            .collect_vec();
```
JonasWanke: Is this line necessary if we only iterate over it?

compiler/src/vm/tracer.rs
```diff
@@ -140,67 +175,226 @@ impl Tracer {
...
-                    args.iter().map(|arg| arg.format(heap)).join(" ")
-                )),
-                TraceEntry::CallEnded { return_value } => {
-                    Action::End(format!(" = {}", return_value.format(heap)))
+                EventData::CallEnded { return_value } => {
+                    let span = stack.pop().unwrap();
+                    let (id, closure, args) = match span.data.unwrap() {
+                        StackEntry::Call { id, closure, args } => (id, closure, args),
+                        _ => unreachable!(),
+                    };
```
JonasWanke:
```suggestion
                    let StackEntry::Call { id, closure, args } = span.data.unwrap()  else { unreachable!(); };
```

compiler/src/vm/tracer.rs
```diff
@@ -140,67 +175,226 @@ impl Tracer {
...
+                EventData::NeedsEnded { nothing } => {
+                    let span = stack.pop().unwrap();
+                    let (id, condition, reason) = match span.data.unwrap() {
+                        StackEntry::Needs {
+                            id,
+                            condition,
+                            reason,
+                        } => (id, condition, reason),
+                        _ => unreachable!(),
+                    };
```
JonasWanke:
```suggestion
                    let StackEntry::Needs { id, condition, reason } = span.data.unwrap() else { unreachable!(); };
```

compiler/src/vm/tracer.rs
```diff
@@ -140,67 +175,226 @@ impl Tracer {
...
+                    });
                 }
-                Action::Stay(line) => {
-                    lines.push(format!("{indent}{line}"));
+                EventData::ModuleEnded { export_map } => {
+                    let span = stack.pop().unwrap();
+                    let module = match span.data.unwrap() {
+                        StackEntry::Module { module } => module,
+                        _ => unreachable!(),
+                    };
```
JonasWanke:
```suggestion
                    let StackEntry::Module { module } = span.data.unwrap() else { unreachable!(); };
```

compiler/src/vm/tracer.rs
```diff
@@ -140,67 +175,226 @@ impl Tracer {
...
+                        start: event.when,
+                        data: Some(StackEntry::Needs {
+                            id: id.clone(),
+                            condition: *condition,
+                            reason: *reason,
+                        }),
+                        inner: vec![],
+                    });
+                }
+                EventData::NeedsEnded { nothing } => {
```
JonasWanke, line 283: Instead of storing `Nothing` for every `::NeedsEnded` just for this purpose, we could store it just once, e.g., in the tracer. However, this might be premature optimization. Also, since `::NeedsEnded` is still smaller than, e.g., `::NeedsStarted`, the enum's size in memory won't change

JonasWanke (changes requested):

compiler/src/vm/heap/mod.rs
MarcelGarus: Tbh, I don't remember why I changed that. I reverted it (and applied clippy lints).

compiler/src/vm/heap/object.rs
MarcelGarus: True. This was mainly so that the `IdGenerator` works, but I changed that to an explicit `CountableId` trait.

compiler/src/vm/mod.rs
MarcelGarus: For the `main` function, we want to be able to call `create_channel` on the VM before calling `set_up_for_running_closure` with the environment that then contains the returned `ChannelId`s.

We could of course just pass an `IdGenerator<ChannelId>` into the constructors as well, would be nicer?

compiler/src/vm/mod.rs
MarcelGarus: I just tried to use it, but couldn't find it. I know that the `im` has a `hash_map!` macro for immutable hash maps. For `std::collection::hash_map::HashMap`, we'd have to use a package such as [`common_macros`](https://docs.rs/common_macros/0.1.1/common_macros/macro.hash_map.html).

Otherwise, I'd update this to the recommended `HashMap::from([some static array])` syntax that's also recommended in the docs.

compiler/src/vm/mod.rs
MarcelGarus: Oh yes, definitely. I initially started by adding code to the heaps to track how often each channel is referenced, but then I thought that was probably a premature optimization. We should definitely keep an eye on this when doing performance optimizations though.

compiler/src/vm/mod.rs
MarcelGarus: Yes, exactly. Take this code:

```
channel = core.channel.create 1
core.parallel { nursery ->
  core.channel.send channel.send nursery
}
escapedNursery = channel.receive
```

MarcelGarus: And `performing_fiber` is always a single – only single fibers every perform sends and receives, parallel and try never do. I should document that somewhere though.

compiler/src/vm/mod.rs
MarcelGarus: Oh, true. Initially, I thought the outer world would be able to send and receive things on-demand and call some kind of `send` or `receive` method directly on the VM, but it probably makes more sense to do everything pull-based (i.e. the outside world can only send stuff into the VM when a fiber listens and otherwise has to take care by itself of queuing up packets to be sent).

compiler/src/vm/mod.rs
MarcelGarus: Theoretically, yes. Practically, we have the same problem that caused us to have an `Option<Status>` in the fuzzer: Rust only exposes a `&mut V` to the `and_modify` closure, so a nice mapping is not possible.

compiler/src/language_server/hints/constant_evaluator.rs
MarcelGarus: Nice, I didn't know about this syntax!

compiler/src/main.rs
MarcelGarus: `core.channel.receive (core.channel.create 1)` would enter this state. I'll update the message to be more descriptive.

compiler/src/vm/tracer.rs
MarcelGarus: That's true. In the next PR that rearchitects the tracing, I fix this problem, so I won't change this here.

compiler/src/vm/mod.rs
JonasWanke: Ah, then `im`'s macro is the one I remembered. In that case, `HashMap::from([…])` sounds good

compiler/src/vm/mod.rs
JonasWanke: Oh, right, makes sense

compiler/src/language_server/hints/constant_evaluator.rs
JonasWanke: It's pretty new, the feature gate just got removed. https://github.com/rust-lang/rust/issues/87335

compiler/src/vm/mod.rs
JonasWanke: Right, I saw that later on but forgot to remove this comment. I'm not sure an `IdGenerator<ChannelId>` would suffice since the VM has to know which channel(s) to create initially. I guess the current way suffices. If VMs are created in more places, we could think about a builder to encapsulate the setup in a type-safe way

compiler/src/vm/fiber.rs
MarcelGarus: Oops right. I actually intended to do that and the Candy code in `Core` relied on that.

compiler/src/main.rs
MarcelGarus: Yes, in the for-loop below we use the `vm` mutably, so we need to buffer the operations locally.

compiler/src/vm/mod.rs
MarcelGarus: On second thought: This might work for external channels, but what happens when a channel is sent to the outside? I added a comment below at the top-level-thread to discuss `Option<FiberId>` and `performing_fiber` in general.

compiler/src/vm/mod.rs
MarcelGarus: Assuming we don't find a way to eliminate using `Option<FiberId>`, this would mean that all 6 call sites of `complete_send` and `complete_receive` would have an `if let` around the call.
I know that these functions are a weak abstraction layer – they only came to be after realizing that the code for performing a send or receive is the same 4 lines in several places. If you think we should really put the `if let` outside the function, I'd rather inline the whole function at the call sites.

compiler/src/vm/mod.rs
MarcelGarus: See a comment further below for more discussion about this.

compiler/src/vm/mod.rs
MarcelGarus: If channels send something to a nursery, the send completes immediately after the new fiber is spawned on the nursery. This call in particular sets the status of the performing fiber back to `Running` so that it's considered as a candidate for further execution in the next `raw_run`.

MarcelGarus:
Some comments are about how the `performing_fiber` relates to the outside world. Tbh, until now, I didn't spend too much time thinking about how the outside world can operate on internal channels. I believe it makes sense to support both pull-based and push-based use cases:

- External channels are managed in the outside world and are pull-based. The outside world can lazily generate values once `Receive` operations happen on those channels.
- Internal channels might reach the outside world over other channels.
  We probably want this. For example, to interact with external services (such as file systems) on a request-response basis, one send and one receive channel are _not_ enough: The send channel might be given to multiple pieces of code that make requests concurrently. They only want to receive responses to their request. A shared response channel makes that impossible –  you'd always be in danger of some other fiber receiving your response.
  For those cases, it instead makes sense to create a new one-off channel for each request and send the send port to the outside. This gives the outside world the opportunity to notify that piece of code specifically.

Essentially, this means that the outside world needs to also have a way of sending and receiving data from internal channels. One idea would be to change the `performing_fiber` that's extensively used in the code from an `Option<FiberId>` to a custom enum that is something like this:

```rust
enum Performer {
  Fiber(FiberId),
  Extern(ExternalOperationId),
  Nursery,
}
```

This way, the outside world could call `receive_from_internal_channel` (or something similar) on a VM to get an `ExternalOperationId`. The VM could then also publish the results of those operations.

I'm not too fond of the name `Performer` by the way. (Maybe `Issuer` because those are the entities that issued the operations? Or `OperationSource`?) If you have suggestions, let me hear them.

compiler/src/vm/mod.rs
MarcelGarus: I also added a comment below at the top-level-thread to discuss `Option<FiberId>` in general.

compiler/src/vm/mod.rs
MarcelGarus: See the top-level comment below

MarcelGarus:
On third thought, maybe it would be even easier if we had no extra concept of external channels at all. Instead, the outside world would always use "internal" channels as well. If we want pull-based behavior, we can still do that using a channel with a capacity of 0.

JonasWanke:
Regarding one send and one receive channel for communication with the outside world: That would work if we add an extra layer inside Candy that takes care of multiplexing and demultiplexing the messages. Basically the job that would otherwise fall to the Rust code.

Using the “internal” channels for communication with the outside world as well sounds good since that should simplify the code while still allowing us to accomplish the same

MarcelGarus:
(Although I already requested a review before, now there's _really_ nothing more I'd change.)
Edit: Except for fixing the Clippy lints, apparently.

compiler/src/vm/mod.rs
```diff
@@ -473,10 +438,13 @@ impl Vm {
...
                 mut paused_fiber,
                 nursery,
+                return_value,
                 ..
             } = tree.into_parallel().unwrap();
             self.channels.remove(&nursery).unwrap();
-            paused_fiber.fiber.complete_parallel_scope(result);
+            paused_fiber
+                .fiber
+                .complete_parallel_scope(result.map(|_| return_value.unwrap()));
```
JonasWanke, line 447: It looks like `result` and `return_value` are partial duplicates. Around line 368, if the fiber completes successfully, the return value is stored in the fiber and then this function is called with `Ok(())`. Can we instead remove `parallel.return_value` and pass the value to this function directly?

compiler/src/vm/mod.rs
```diff
@@ -499,17 +467,13 @@ impl Vm {
...
-        channel_id: ChannelId,
-        packet: Packet,
-    ) {
+    fn send_to_channel(&mut self, performer: Performer, channel_id: ChannelId, packet: Packet) {
         let channel = match self.channels.get_mut(&channel_id) {
             Some(channel) => channel,
             None => {
                 // The channel was a nursery that died.
-                if let Some(fiber) = performing_fiber {
+                // TODO: also panic a nursery performer
```
JonasWanke: Should this be addressed before merging?

compiler/src/main.rs
```diff
@@ -258,45 +262,31 @@ fn run(options: CandyRunOptions) {
...
-                    info!("Sent to stdout: {}", packet.value.format(&packet.heap));
-                    vm.complete_send(performing_fiber);
-                }
-                vm::Operation::Receive { .. } => unreachable!(),
-                vm::Operation::Drop => vm.free_channel(stdout),
+        stdout.run(&mut vm);
+        for channel in vm.unreferenced_channels.iter().copied().collect_vec() {
+            if channel != stdout.channel {
+                vm.free_channel(channel);
             }
```
JonasWanke: `stdout` should behave like other channels and once it's no longer referenced, it won't be passed to the VM again.

Also, I'm not sure whether `.collect_vec()` is needed for the `for` loop's iterable

packages/Core/concurrency.candy
```diff
@@ -0,0 +1,18 @@
+channel = use "..channel"
+function = use "..function"
+structGetUnwrap = (use "..struct").getUnwrap
```
JonasWanke:
```suggestion
```

JonasWanke (changes requested):

compiler/src/main.rs
MarcelGarus: Fair enough. The `collect_vec` is actually needed though because the `free_channel` method needs a mutable VM. It also removes the channel from the `unreferenced_channels` (which would not be a good idea to do while we are iterating over that).

compiler/src/vm/mod.rs
MarcelGarus: I first thought about doing that, but it may be that the first fiber exits way before the last fiber of the parallel section.

```candy
core.parallel { nursery ->
  core.async {
    # long running computation
  }
  Foo
}
```
 
In this example, the initial child returns `Foo` while the spawned fiber is still running.

compiler/src/vm/mod.rs
MarcelGarus: Oh, that todo is already implemented by the code below. I just forgot to delete it.

compiler/src/vm/mod.rs
JonasWanke: Oh, that makes sense

JonasWanke (approved):

PR #243, opened 2022-10-09
Doc comments

JonasWanke:


compiler/src/compiler/comment/string_to_rcst.rs
```diff
@@ -0,0 +1,1500 @@
...
+        .filter_map(|it| match &it.kind {
+            cst::CstKind::Comment { comment, .. } => {
+                Some(comment.strip_prefix(' ').unwrap_or(comment))
+            }
+            cst::CstKind::Newline(_) => None,
+            _ => unreachable!(),
+        })
+        .collect_vec();
+
+    let (rest, rcsts) = parse::blocks(comment_lines, 0).unwrap();
```
MarcelGarus: In the parsers below, `rest` is called `remaining`, which I like more. Do you think that would fit here as well?

MarcelGarus (changes requested):
Nice work! The parser is surprisingly readable.

MarcelGarus (approved):

PR #244, opened 2022-10-17
Tracing

MarcelGarus:
Depends on #241 

#241 changed the VM architecture a lot. In particular, it introduced the ability to run several concurrently running fibers. This broke tracing, which still relied on some assumptions such as all objects being in a single heap.

This PR fixes the tracing by moving the tracer outside of the fiber's and VM's ownership. Instead, the tracer is borrowed to the `vm.run` method, along with the other existing context parts passed to it. Depending on the amount of insight that the outside world wants to get into the execution, it can pass in different tracers. The following tracers exist for now:

- **Dummy tracer:** Saves nothing.
- **Fuzzable closures finder:** Only tracks all found fuzzable closures.
- **Full tracer:** Saves every single event that occurs.

We can use the full tracer to figure out exactly what happened in the VM. Using this, we can correctly output which part of the code is responsible and we have nice stack traces, even across fibers.

```candy
core = use "..Core"

main environment :=
  print message = { core.channel.send environment.stdout message }
  print "Hello, world!"

  core.parallel { nursery ->
    print (core.await (core.async nursery {
      print "Hello from fiber!"
      "Hello, async await!"
    }))

    core.async nursery { print "Kiwi" }
    core.async nursery { core.panic "life's not fair" }
    print "Hi"
  }
```

beautiful stack trace

Previously, we also supported a hierarchical trace dump that related calls to returns, etc. For now, in the new architecture, we only have a straightforward list of all events. However, I plan on introducing a Flamegraph-inspired graphical trace visualization that looks something like this:

potential tracing visualization

This would allow us to do profiling of the Candy code itself and also see dependencies between fibers.


compiler/src/fuzzer/utils.rs
```diff
@@ -1,46 +1,72 @@
...
+}
+impl FuzzablesFinder {
+    pub fn new() -> Self {
+        Self {
+            fuzzables: vec![],
+            heap: Heap::default(),
+            transferred_objects: HashMap::new(),
+        }
+    }
+}
```
JonasWanke: Maybe derive `Default`?

compiler/src/vm/heap/mod.rs
```diff
@@ -125,50 +128,54 @@ impl Heap {
 
     /// Clones all objects at the `root_addresses` into the `other` heap and
     /// returns a list of their addresses in the other heap.
-    pub fn clone_multiple_to_other_heap(
+    pub fn clone_multiple_to_other_heap_with_existing_mapping(
         &self,
         other: &mut Heap,
         addresses: &[Pointer],
+        address_map: &mut HashMap<Pointer, Pointer>,
```
JonasWanke, line 135: IIUC, passing the existing mappings to avoid duplicate copies will break when we add GC since the value behind a given address could have changed by then

compiler/src/language_server/hints/constant_evaluator.rs
```diff
@@ -62,31 +68,38 @@ impl ConstantEvaluator {
...
         let span = span!(Level::DEBUG, "Calculating hints for {module}");
         let _enter = span.enter();
 
-        let vm = &self.vms[module];
+        let evaluator = &self.evaluators[module];
         let mut hints = vec![];
 
-        if let vm::Status::Panicked { reason } = vm.status() {
-            if let Some(hint) = panic_hint(db, module.clone(), vm, reason) {
+        // TODO: Think about how to highlight the responsible piece of code.
```
JonasWanke, line 97: We could also publish [LSP diagnostics](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic) similar to parser errors so that the caller gets highlighted in red and then reference the failed `needs` or maybe even the whole call chain using the `relatedInformation` field

compiler/src/main.rs
```diff
@@ -270,13 +273,18 @@ fn run(options: CandyRunOptions) {
         let stdout_port = heap.create_send_port(stdout.channel);
         heap.create_struct(HashMap::from([(stdout_symbol, stdout_port)]))
     };
+    tracer.in_fiber_tracer(FiberId::root()).call_started(
+        &heap,
+        Id::new(module, vec!["main".to_string()]),
+        main,
+        vec![environment],
+    );
```
JonasWanke: Shouldn't this call be the responsibility of the VM, maybe in `set_up_for_running_closure(…)`? The same goes for the tracing of `call_ended(…)` below

compiler/src/vm/tracer/stack_trace.rs
```diff
@@ -0,0 +1,184 @@
...
+    Module {
+        module: Module,
+    },
+}
+
+impl FullTracer {
+    pub fn stack_traces(&self) -> HashMap<FiberId, Vec<StackEntry>> {
+        let mut stacks: HashMap<FiberId, Vec<StackEntry>> = HashMap::new();
+        for timed_event in &self.events {
+            if let Event::InFiber { fiber, event } = &timed_event.event {
```
JonasWanke: You can remove one level of indentation:

```suggestion
            let Event::InFiber { fiber, event } = &timed_event.event else { continue; };
```

compiler/src/vm/tracer/stack_trace.rs
```diff
@@ -0,0 +1,184 @@
...
+                let stack = stacks.entry(*fiber).or_default();
+                match event {
+                    InFiberEvent::ModuleStarted { module } => {
+                        stack.push(StackEntry::Module {
+                            module: module.clone(),
+                        });
+                    }
+                    InFiberEvent::ModuleEnded { .. } => {
+                        stack.pop().unwrap();
```
JonasWanke: The `…Ended` cases could assert that the popped entry is of the same type, but it's not necessary

compiler/src/vm/tracer/stack_trace.rs
```diff
@@ -0,0 +1,184 @@
...
+                StackEntry::Needs {
+                    condition, reason, ..
+                } => format!(
+                    "needs {} {}",
+                    condition.format(&self.heap),
+                    reason.format(&self.heap),
+                ),
+                StackEntry::Module { module } => format!("module {module}"),
+            };
+            lines.push(format!("{caller_location_string:90} {call_string}"));
```
JonasWanke: Instead of this hard-coded alignment, can we check how wide the location string actually becomes in the longest line?

compiler/src/vm/tracer/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+    fn needs_ended(&mut self);
+}
+
+// A dummy version of the tracer that is used when running known instructions
+// without wanting to trace them.
+
+pub struct DummyTracer;
+pub struct DummyInFiberTracer;
+impl Tracer for DummyTracer {
+    fn fiber_created(&mut self, _fiber: FiberId) {}
```
JonasWanke: Since we have the `Event` enum anyway, maybe we should simplify the `Tracer` to just have a single method accepting an `Event`? That should save quite a few lines in the `Tracer` implementations

compiler/src/vm/fiber.rs
```diff
@@ -104,32 +111,43 @@ impl Fiber {
...
         );
         fiber
     }
     pub fn new_for_running_module_closure(closure: Closure) -> Self {
-        assert_eq!(closure.captured.len(), 0, "Called start_module_closure with a closure that is not a module closure (it captures stuff).");
-        assert_eq!(closure.num_args, 0, "Called start_module_closure with a closure that is not a module closure (it has arguments).");
+        assert_eq!(
+            closure.captured.len(),
+            0,
+            "closure is not a module closure (it captures stuff)."
```
```suggestion
            "Closure is not a module closure (it captures stuff)."
```

compiler/src/vm/fiber.rs
```diff
@@ -104,32 +111,43 @@ impl Fiber {
...
+        );
+        assert_eq!(
+            closure.num_args, 0,
+            "closure is not a module closure (it has arguments)."
```
```suggestion
            "Closure is not a module closure (it has arguments)."
```

compiler/src/vm/fiber.rs
```diff
@@ -177,19 +195,21 @@ impl Fiber {
...
-                    .clone_single_to_other_heap(&mut self.heap, packet.value);
+                    .clone_single_to_other_heap(&mut self.heap, packet.address);
                 self.data_stack.push(value);
                 self.status = Status::Running;
             }
             Err(reason) => self.panic(reason),
         }
     }
-    pub fn complete_try(&mut self, result: Result<Packet, String>) {
-        assert!(matches!(self.status, Status::InTry { .. }));
```
JonasWanke, line 188: Why did you remove this assertion?

compiler/src/vm/context.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
 // of the VM.
 
 pub trait UseProvider {
-    fn use_module(&self, module: Module) -> Result<UseResult, String>;
+    fn use_module(&mut self, module: Module) -> Result<UseResult, String>;
```
JonasWanke: Why does this have to receive a mutable self?

compiler/src/vm/heap/mod.rs
MarcelGarus: Oh, right, I didn't think of that.

We'll definitely need to look into that then. We could of course look into using a deduplicating heap for the tracing so that even though the objects are fully traversed every time, only one instance is stored by the tracer. But I think it's fine that we do that later when we introduce a better heap representation than a `HashMap`.

Do you agree or do you think I should address this now?

compiler/src/language_server/hints/constant_evaluator.rs
MarcelGarus: I like that. Do you think it's fine if we look at that in a separate PR?

JonasWanke: Yes, please :)

compiler/src/vm/tracer/mod.rs
MarcelGarus: Hmm, maybe. I definitely agree that there is some boilerplate code when implementing a tracer. But when designing the tracer, I first thought about what an optimal usage API would look like and then implemented that.

I feel like while taking an `Event` makes the tracer implementations themselves cleaner, it pushes some complexity into the call sites. For example, _some_ methods also take a `Heap` so they can clone objects into the tracer. If we create a general method that accepts an `Event`, that would also need to accept a heap.

I actually started doing the work for it last night, but there are some ugly changes to the caller locations:

```rust
tracer.fiber_execution_started(fiber_id);
// becomes:
tracer.add(
  Event::FiberExecutionStarted { fiber: fiber_id },
  &Heap::new(),
);
```

Actually, now that I think about it, what do you think about splitting the tracer further up into a `Tracer` and `InFiberTracer` with an optimal API and a `TracerImplementation` which only has a single method accepting an `Event`? This way, we would have both a clean API as well as simple implementations.

compiler/src/vm/heap/mod.rs
JonasWanke: No, we can address this later when it becomes relevant. Deduplication in general sounds good since it could also improve fiber communication, but, in this case, we could also check whether the value is the same as we previously copied from that address or we could add a generation counter for the values behind an address. But we can decide on that later on

compiler/src/vm/tracer/mod.rs
```diff
@@ -103,241 +30,155 @@ pub enum Event {
...
         }
     }
-    fn push(&mut self, data: Event) {
-        self.events.push(TimedEvent {
-            when: Instant::now(),
-            event: data,
-        });
+}
+
+impl<'vm> VmTracer<'vm> {
```
JonasWanke:
```suggestion
impl<'vm> Tracer for VmTracer<'vm> {
```

compiler/src/vm/tracer/mod.rs
```diff
@@ -103,241 +30,155 @@ pub enum Event {
...
+            fiber: self.fiber,
+            event,
+        });
     }
 }
-impl Tracer for FullTracer {
-    fn fiber_created(&mut self, fiber: FiberId) {
-        self.push(Event::FiberCreated { fiber });
+
+impl<'vm> VmTracer<'vm> {
```
JonasWanke:
```suggestion
impl Tracer {
```

compiler/src/vm/tracer/full.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+}
+
+impl fmt::Debug for FullTracer {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let start = self.events.first().map(|event| event.when);
+        for event in &self.events {
+            writeln!(
+                f,
+                "{:?} us: {}",
```
JonasWanke:
```suggestion
                "{:?} µs: {}",
```

JonasWanke (approved):

JonasWanke (approved):

PR #245, opened 2022-10-17
Add guidelines for writing sensible needs reasons

MarcelGarus:
Depends on #244 

Adds guidelines for reasons and applies them in the whole project. Also fixes some problems that were found by fuzzing.

PR #246, opened 2022-10-19
Add fuzzing to the CI

MarcelGarus:
Depends on #245

This makes the exit code of the `candy fuzz` command match whether errors were found during fuzzing. It also adds fuzzing to the CI and fixes some errors that were discovered.


MarcelGarus:
In the logs, you may have noticed that the functions in `Core` are fuzzed several times. That's because the tree of modules created from `Core/_.candy` contains the outer leaf files many times. For example, the `core.bool` module is imported by many other modules in the core package, causing it to be instantiated and evaluated many times. All the created closures are separate and fuzzed independently.

I guess I'll start working on some basic optimizations such as function inlining, duplicate subtree elimination, and compile-time evaluation soon to fix the problem described above. LLVM would also do those optimizations, but we'd have to turn the entire tree (including expanded uses, thus an exponential amount of code) into LLVM-IR for it to do its magic, so that's too late in the pipeline and it makes sense to do these optimizations in Candy itself.

packages/Benchmark.candy
```diff
@@ -3,13 +3,14 @@
...
-fibRec fibRec n =
+fibRec = { fibRec n ->
   core.ifElse (core.int.isLessThan n 2) { n } {
     core.int.add
       fibRec fibRec (core.int.subtract n 1)
       fibRec fibRec (core.int.subtract n 2)
   }
-fib n = fibRec fibRec n
+}
+fib = { n -> fibRec fibRec n }
```
JonasWanke:
```suggestion
fib n =
  needs (core.int.is n)
  fibRec fibRec n
```

packages/Core/int.candy
```diff
@@ -73,6 +73,13 @@ absolute value :=
   needs (is value)
   conditionals.ifElse (isNegative value) (negate value) value
 
+fitsInRustU32 value =
```
JonasWanke:
```suggestion
fitsInRustU32 value :=
```

.github/workflows/compiler.yaml
```diff
@@ -67,3 +67,19 @@ jobs:
...
+    name: Fuzzing
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ env.RUST_VERSION }}
+          override: true
+          components: clippy
```
JonasWanke: I don't think we need Clippy for fuzzing

```suggestion
```

compiler/src/main.rs
```diff
@@ -97,15 +97,28 @@ async fn main() {
...
     );
-    raw_build(module.clone(), options.debug);
+    let result = raw_build(module.clone(), options.debug);
 
-    if options.watch {
+    if !options.watch {
+        match result {
+            Some(_) => Ok(()),
+            None => Err(Exit::FileNotFound),
+        }
```
JonasWanke: Blub

JonasWanke (changes requested):
Rust also has a few complaints:

image

MarcelGarus:
Fuzzing works fine locally. Seems like this is an issue of nightly cargo: https://github.com/rust-lang/cargo/issues/8557

JonasWanke (commented):
I'm not sure whether the issue you linked is really related since it's been closed for more than two years, but I couldn't find a good-looking current one. Maybe we should disable the action for now so that we don't get used to the failing CI?

Other than that, the PR looks fine

MarcelGarus:
Fair enough, I'll just disable the fuzzing for now

JonasWanke (approved):

PR #247, opened 2022-11-03
List literals

JonasWanke:
image

Partially reverts #234

### Checklist


compiler/src/compiler/string_to_rcst.rs
```diff
@@ -1059,6 +1060,251 @@ mod parse {
...
+                        value: Box::new(Rcst::Error {
+                            unparsable_input: "".to_string(),
+                            error: RcstError::ListItemMissesValue,
+                        }),
+                        comma: Some(Box::new(Rcst::Comma)),
+                    }],
+                    closing_parenthesis: Box::new(Rcst::ClosingParenthesis),
+                },
+            )),
+        );
```
MarcelGarus: Isn't this supposed to be an empty list?

compiler/src/compiler/string_to_rcst.rs
```diff
@@ -1059,6 +1060,251 @@ mod parse {
...
+                    ],
+                    closing_parenthesis: Box::new(Rcst::ClosingParenthesis),
+                },
+            )),
+        );
+        // [
+        //   foo,
+        //   4,
+        //   "Hi",
+        // ]
```
MarcelGarus:
```suggestion
        // (
        //   foo,
        //   4,
        //   "Hi",
        // )
```

compiler/src/vm/builtin_functions.rs
```diff
@@ -292,6 +298,59 @@ impl Heap {
...
+    fn list_create(&mut self, args: &[Pointer]) -> BuiltinResult {
+        unpack_and_later_drop!(self, args, |length: Int| {
+            let length = length.value.to_usize().unwrap();
+            let nothing = self.create_nothing();
+            Return(self.create_list(vec![nothing; length]))
+        })
+    }
+    fn list_get(&mut self, args: &[Pointer]) -> BuiltinResult {
+        unpack_and_later_drop!(self, args, |list: List, index: Int| {
+            let index = index.value.to_usize().unwrap();
```
MarcelGarus, line 309: If `builtinListGet` is called with a negative number, we probably don't want to crash the entire VM/LSP with `unwrap`. Instead, we probably want to do something like this:

```suggestion
            let index = index.value.to_usize().map_err(|_| "builtinListGet needs a non-negative index")?;
```

The same applies to all the argument handling in this file.

compiler/src/vm/heap/object.rs
```diff
@@ -235,6 +259,10 @@ impl Data {
             Data::Int(int) => format!("{}", int.value),
             Data::Text(text) => format!("\"{}\"", text.value),
             Data::Symbol(symbol) => symbol.value.to_string(),
+            Data::List(List { items }) => format!(
+                "({})",
+                items.iter().map(|item| item.format(heap)).join(", ")
+            ),
```
MarcelGarus: AFAICT this is incorrect for empty lists

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #248, opened 2022-11-10
Pipe operator and Core updates

JonasWanke:
**This PR depends on #247!**

### Checklist


packages/Core/channel.candy
```diff
@@ -9,14 +9,14 @@ isReceivePort value := isType value ReceivePort
 
 create capacity :=
   needs (isInt capacity) "`capacity` is not an integer. Channels need a capacity because otherwise, there would be no backpressure among fibers, and memory leaks would go unnoticed."
-  ✨.channelCreate capacity
+  capacity | ✨.channelCreate
```
MarcelGarus: Sure about this change? Even in object-oriented languages, I'd usually not do `3.createChannel()` but rather `createChannel(3)`

packages/Core/concurrency.candy
```diff
@@ -4,15 +4,15 @@ if = (use "..conditionals").if
 
 parallel body :=
   needs (function.is1 body) "The `body` should be a function taking a nursery."
-  ✨.parallel body
+  body | ✨.parallel
```
MarcelGarus: Same here. Not sure I'd use the pipe here

packages/Core/list.candy
```diff
@@ -1,70 +1,75 @@
...
 int = use "..int"
 type = use "..type"
 
-create length :=
+filled length item :=
   needs (int.is length)
   needs (int.isNonNegative length)
   needs (int.fitsInRustU32 length)
-  ✨.listCreate length
+  length | ✨.listFilled item
```
MarcelGarus: Not sure bout this

packages/Core/text.candy
```diff
@@ -9,21 +9,21 @@ is value := type.is value Text
...
   needs (is text)
-  ✨.textCharacters text
+  text | ✨.textCharacters
 characterAt text index :=
   needs (is text)
   needs (int.is index)
   needs (int.isNonNegative index)
   needs (int.isLessThan index (length text))
-  struct.getUnwrap (characters text) index
+  text | characters | struct.getUnwrap index
```
MarcelGarus: One could argue that `✨.textCharacters` should return a list instead of a struct

old-packages/Core/src/Todo.candy
```diff
@@ -1,9 +0,0 @@
-use ..Panic
-use ..Primitives
-use ..String
-
-public fun todo(message: String = "Not implemented"): Never {
-  # Always panics with a todo message.
```
MarcelGarus, line 6: I'd also port the doc comment

old-packages/Core/src/Collections/List/.candy
```diff
@@ -8,61 +8,14 @@ use ..Iterate
 public use .Array
 
 public trait List[Item]: Iterate[Item] {
-  public static fun filled(length: Int /* UInt */, item: Item): List[Item] {
-    Array[Item].filled(length, item)
-  }
   public static fun generate(length: Int, generator: (Int) -> Item): List[Item] {
     Array[Item].generate(length, generator)
   }
-  public static fun empty(): List[Item] { Array[Item].empty() }
```
MarcelGarus, line 17: Just to make sure: We decided to not have a `list.empty` in favor of the literal, right?

old-packages/Core/src/Assert.candy
```diff
@@ -1,12 +0,0 @@
-use ..Bool
-use ..Panic
-use ..String
-
-public fun assert(condition: Bool, message: String = "Assert failed.") {
-  # Throws an error if the `condition` is `false`.
-  #
-  # This function is useful for ensuring that your mental model of the state of your program matches
-  # its actual state.
```
MarcelGarus, line 9: I'd also port the doc comment to the `core.check` function

MarcelGarus (changes requested):

old-packages/Core/src/Collections/List/.candy
JonasWanke: Correct

packages/Core/text.candy
JonasWanke: It actually did already since it used the `create_list(…)` function that changed from creating our struct-based list representation to a proper list, but this place wasn't updated. I agree that a list makes more sense for now, but we should probably switch to an iterable, at least when we have our own Unicode implementation in Candy

JonasWanke:
Sorry for the superfluous pipe usages; I was using search+replace and apparently not being careful enough about where it actually makes sense

MarcelGarus (approved):

PR #255, opened 2022-12-02
Iterable

JonasWanke:
### Checklist


packages/Core/iterable.candy
```diff
@@ -0,0 +1,167 @@
...
+    newItem | result.map { newItem ->
+      (newItem, rawMapWithStateHelper newState mapper)
+    }
+  }
+}
+rawMapWithState initialState mapper :=
+  # `mapper` receives the previous state and returns the optional item to emit
+  # (or none if it's done) and the new state.
+  needs (function.is1 mapper)
+  rawMapWithStateHelper rawMapWithStateHelper initialState mapper
```
MarcelGarus: I'd drop the `raw` prefix or make this non-public or both.

packages/Core/iterable.candy
```diff
@@ -0,0 +1,167 @@
...
+sum iterableOfInt :=
+  needs (is iterableOfInt)
+  iterableOfInt | reduceLeft { a b ->
+    needs (int.is a)
+    needs (int.is b)
+    a | int.add b
+  }
+length iterable :=
+  needs (is iterable)
+  iterable | map { item -> 1 } | sum | result.unwrapOr 0
```
MarcelGarus:
```suggestion
  iterable | map { _ -> 1 } | sum | result.unwrapOr 0
```

packages/Core/list.candy
```diff
@@ -22,24 +24,30 @@ isEmpty list :=
 
 lastIndex list :=
   needs (is list)
-  needs (bool.not (isEmpty list))
-  length list | int.minus 1
+  ifElse
+    (list | isEmpty)
+    result.error "Can't get the last index of an empty list."
+    list | length | int.minus 1 | result.ok
```
MarcelGarus:
```suggestion
    { result.error "Can't get the last index of an empty list." }
    { list | length | int.minus 1 | result.ok }
```

packages/Core/list.candy
```diff
@@ -22,24 +24,30 @@ isEmpty list :=
...
   needs (is list)
-  int.is index | bool.lazyAnd {
-    int.isNonNegative index
-    | bool.and (index | int.isLessThanOrEqualTo (lastIndex list))
-  }
+  needs (int.is index)
+  ifElse
+    (list | isEmpty)
+    False
+    index | int.isNonNegative
```
MarcelGarus:
```suggestion
    { False }
    { index | int.isNonNegative }
```

MarcelGarus (changes requested):

packages/Core/iterable.candy
MarcelGarus: Also, this doesn't really map iterables anything, right? For example, our implementation of a list-to-iterable or range-to-iterable converters could also use this. So, maybe call this `newWithState` or `newFromState` instead to highlight it creates a new iterable from a state and mapper?

packages/Core/iterable.candy
```diff
@@ -0,0 +1,167 @@
...
+# where iterable tester :=
+#   needs (is iterable)
+#   needs (function.is1 tester)
+
+## Action
+
+forEach iterable action :=
+  needs (is iterable)
+  needs (function.is1 action)
+  iterable | map action
```
MarcelGarus: This won't behave as expected because `map` is lazy, while `forEach` shouldn't be. For example, `("Foo", "Bar") | listToIterable | forEach { item -> print item }` would just return a closure rather than printing the items.

Maybe change it into this?

```suggestion
forEach iterable action :=
  needs (is iterable)
  needs (function.is1 action)
  iterable
  | map { item ->
    action item
    Nothing
  }
  | toList
```

MarcelGarus (changes requested):

packages/Core/iterable.candy
```diff
@@ -0,0 +1,167 @@
+bool = use "..bool"
```
MarcelGarus, line 1: I believe this file should be exported in `_.candy`.

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #256, opened 2022-12-02
Implement Tail Call Optimization

MarcelGarus:
This PR adds support for tail calls into the compiler (LIR and VM).


compiler/src/compiler/mir_optimize/module_folding.rs
```diff
@@ -52,11 +52,11 @@ impl Mir {
                     return; // TODO: Replace with a panic.
                 };
                 let Ok(path) = UsePath::parse(path) else {
-                    warn!("`use` called with an invalid path.");
+                    warn!("`use` called with an invalid path: `\"{path}\"`");
```
JonasWanke:
```suggestion
                    warn!("`use` called with an invalid path: `\"{path}\"`.");
```

compiler/src/compiler/mir_optimize/module_folding.rs
```diff
@@ -52,11 +52,11 @@ impl Mir {
...
                     return; // TODO: Replace with a panic.
                 };
                 let Ok(module_to_import) = path.resolve_relative_to(current_module.clone()) else {
-                    warn!("`use` called with an invalid path.");
+                    warn!("`use` called with a path that doesn't refer to a module ({path:?} relative to {current_module}).");
```
JonasWanke: For consistency

```suggestion
                    warn!("`use` called with a path that doesn't refer to a module: `\"{path:?}\"` relative to {current_module}.");
```

compiler/src/vm/builtin_functions.rs
```diff
@@ -400,7 +387,7 @@ impl Heap {
                 }
                 None => Err(format!(
                     "The struct does not contain the key {}.",
-                    key.format(self)
+                    key.address.format(self)
```
JonasWanke:
```suggestion
                    key.address.format(self),
```

compiler/src/vm/fiber.rs
```diff
@@ -107,17 +106,7 @@ impl Fiber {
...
-        fiber.data_stack.extend(arguments);
-        fiber.data_stack.push(responsible);
-        fiber.run_instruction(
-            &PanickingUseProvider,
-            &mut DummyTracer.for_fiber(FiberId::root()),
-            Instruction::Call {
-                num_args: arguments.len(),
-            },
-        );
+        fiber.call(closure, arguments.to_vec(), responsible);
```
JonasWanke: Maybe take the `arguments` as a `Vec<Pointer>` directly?

compiler/src/vm/fiber.rs
```diff
@@ -500,6 +475,47 @@ impl Fiber {
...
+    pub fn call(&mut self, callee: Pointer, arguments: Vec<Pointer>, responsible: Pointer) {
+        match &self.heap.get(callee).data {
+            Data::Closure(Closure {
+                captured,
+                num_args: expected_num_args,
+                ..
+            }) => {
+                if arguments.len() != *expected_num_args {
+                    self.panic(
+                        format!("A closure expected {expected_num_args} parameters, but you called it with {} arguments.", arguments.len()),
```
```suggestion
                        format!(
                            "A closure expected {expected_num_args} parameters, but you called it with {} arguments.",
                            arguments.len(),
                        ),
```

compiler/src/vm/fiber.rs
```diff
@@ -500,6 +475,47 @@ impl Fiber {
...
+                        self.heap.get_hir_id(responsible),
+                    );
+                    return;
+                }
+
+                self.call_stack.push(self.next_instruction);
+                self.data_stack.append(&mut captured.clone());
+                for captured in captured.clone() {
+                    self.heap.dup(captured);
+                }
```
JonasWanke: `vec::append(…)` empties the other `Vec`: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append. Hence, this would not `dup(…)` any objects right now

```suggestion
                for captured in captured.clone() {
                    self.heap.dup(captured);
                }
                self.data_stack.append(&mut captured.clone());
```

packages/Core/result.candy
```diff
@@ -1,7 +1,7 @@
 bool = use "..bool"
-conditionals = use "..conditionals"
 equals = (use "..equality").equals
 function = use "..function"
+ifElse = (use "..controlFlow").ifElse
```
JonasWanke: I'd keep these ordered by import path

packages/Core/int.candy
```diff
@@ -1,7 +1,7 @@
 bool = use "..bool"
 check = (use "..check").check
-conditionals = use "..conditionals"
 equals = (use "..equality").equals
+ifElse = (use "..controlFlow").ifElse
```
JonasWanke, line 3: I'd keep these ordered by import path

JonasWanke (changes requested):

JonasWanke (approved):

PR #257, opened 2022-12-10
Destructuring

JonasWanke:
image

The duplicate `#` in hints and their order are not looking good, but I want to try using LSP inline values for them instead: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlineValue.

This PR also fixes the semantic tokens for identifiers by passing down the context-inferred type and only marking them as functions when they are actually called.

### Checklist


MarcelGarus (approved):

MarcelGarus:
While meeting yesterday, we also decided that the Candy code should be updated to use pattern matching where appropriate.

packages/Core/controlFlow.candy
```diff
@@ -1,7 +1,7 @@
 bool = use "..bool"
-equals = (use "..equality").equals
+[equals] = use "..equality"
 function = use "..function"
-isSymbol = (use "..symbol").is
+symbol = use "..symbol"
```
MarcelGarus: I was confused why this didn’t cause code below to change. Turns out, the symbol module is never used. So I guess you can also remove all of this

MarcelGarus (changes requested):

MarcelGarus (approved):

language.md
```diff
@@ -134,6 +137,20 @@ Green
 Foo
 ```
 
+### Lists
+
+A list is a compound object that holds aa collection of elements identified by integer index.
```
MarcelGarus:
```suggestion
A list is a compound object that holds a collection of elements identified by integer index.
```

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #267, opened 2023-01-12
Fuzz the compiler itself

MarcelGarus:
This PR adds fuzzing to the compiler itself.
It doesn't yet fix any cases found through fuzzing.


compiler/fuzz/.gitignore
```diff
@@ -0,0 +1,4 @@
+target
+corpus
+artifacts
+coverage
```
JonasWanke: Maybe we should order these alphabetically?

compiler/fuzz/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+name = "candy-fuzz"
+version = "0.0.0"
+publish = false
+edition = "2021"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+candy = { path=".." }
```
JonasWanke:
```suggestion
candy = { path = ".." }
```

compiler/fuzz/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+lazy_static = "1.4.0"
+libfuzzer-sys = "0.4"
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
```
JonasWanke: Could we lift this up to the parent workspace so that the build cache is also shared?

compiler/fuzz/fuzz_targets/fuzz_parser.rs
```diff
@@ -0,0 +1,93 @@
...
+
+fuzz_target!(|data: &[u8]| {
+    let Ok(source_code) = String::from_utf8(data.to_vec()) else {
+        return;
+    };
+
+    let module_provider = InMemoryModuleProvider::for_modules::<String>(HashMap::new());
+    let mut db = Database::new(Box::new(module_provider));
+
+    db.did_open_module(&MODULE, source_code.as_bytes().to_owned());
```
JonasWanke:
```suggestion
    let mut module_provider = InMemoryModuleProvider::default();
    module_provider.add(&MODULE, data.to_vec());
    let db = Database::new(Box::new(module_provider));

```

This version is a bit shorter, more explicit, and also tests our parser against invalid UTF-8 (which, BTW, causes our compiler to crash even with a single byte)

JonasWanke (changes requested):
Can you add a brief README with instructions on how to run the fuzzer and maybe a link to the fuzz book? It looks one has to install the fuzzer first (`cargo install cargo-fuzz`; requires Rust nightly) and can then run a target using `cargo fuzz run fuzz_parser`

JonasWanke (approved):

JonasWanke:
The Clippy complaints are fixed in #268 (https://github.com/candy-lang/candy/pull/268/commits/04c2ead107c24765a058abd9a839b2d57a2b3430). We can wait for that PR to get merged and then update this one, or just merge this one since the complaints are only about a renaming

PR #273, opened 2023-01-19
Fix several errors found using fuzzing

MarcelGarus:
In particular, these cases no longer crash the compiler:

- Assignments with an invalid left-hand side such as a lambda
- Parentheses in patterns
- Patterns that capture no identifiers
- RCST errors in assignment names
- RCST errors in texts
- Unexpected whitespace such as tabs in string indentation
- Pipes in patterns
- Calls in patterns

Some of these are errors, but they just propagate to the next compiler stage instead of crashing the entire compiler.


compiler/src/compiler/string_to_rcst.rs
```diff
@@ -566,6 +566,10 @@ mod parse {
             whitespaces_and_newlines("\nfoo", 0, true),
             ("foo", vec![Rcst::Newline("\n".to_string())])
         );
+        assert_eq!(
+            whitespaces_and_newlines("\nfoo", 1, true),
+            ("\nfoo", vec![])
```
JonasWanke:
```suggestion
            ("\nfoo", vec![]),
```

compiler/src/compiler/error.rs
```diff
@@ -76,6 +80,7 @@ impl Display for CompilerErrorPayload {
                 AstError::ListWithoutClosingParenthesis => {
                     "This list doesn't have a closing parenthesis."
                 }
+                AstError::ParenthesizedInPattern => "You can not use parentheses in patterns.",
```
JonasWanke: For consistency with `AstError::PipeInPattern`:

```suggestion
                AstError::ParenthesizedInPattern => "Parentheses are not allowed in patterns.",
```

Maybe we should also mention that they're fine for list literals? But we could also do that once we add detailed descriptions for the error messages

JonasWanke (changes requested):

JonasWanke (approved):

JonasWanke (approved):

PR #274, opened 2023-01-22
Rewriting the fuzzer

MarcelGarus:
A rewrite of the fuzzer that is more effective because of two things:

- The fuzzer now generates symbols that are actually used in the code. These symbols are found by collecting all symbols of the heap containing only the fuzzed function and its dependencies.
- The fuzzer maintains a pool of inputs and scores them by how many instructions they execute. Inputs that result in long runs (aka inputs that go deep into the function) are favored and mutated.

For example, for the `or` function, the fuzzer is more likely to try symbols (and in particular `True` and `False`) as inputs because those cause more code to be executed:

image


compiler/src/compiler/mir_optimize/module_stack_cancelling.rs
```diff
@@ -38,4 +38,24 @@ impl Mir {
             }
         });
     }
+
+    pub fn remove_all_module_expressions_if_no_use_exists(&mut self) {
+        let mut contains_use = false;
+        self.body.visit(&mut |_, expression, _| {
+            if matches!(expression, Expression::UseModule { .. }) {
+                contains_use = true;
+            }
```
JonasWanke: As a performance improvement, we should probably break this loop early as soon as we detect a remaining `use`. Can you either implement that now or add a comment or a task in the performance epic?

compiler/src/compiler/mir_optimize/module_stack_cancelling.rs
```diff
@@ -38,4 +38,24 @@ impl Mir {
...
+        });
+
+        if !contains_use {
+            self.body.visit(&mut |_, expression, _| {
+                if matches!(
+                    expression,
+                    Expression::ModuleStarts { .. } | Expression::ModuleEnds
```
JonasWanke:
```suggestion
                    Expression::ModuleStarts { .. } | Expression::ModuleEnds,
```

compiler/src/fuzzer/utils.rs
```diff
@@ -1,13 +1,45 @@
...
+}
+impl fmt::Display for Input {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            self.arguments
+                .iter()
+                .map(|arg| format!("{arg:?}"))
+                .join(" ")
```
JonasWanke:
```suggestion
                .join(" "),
```

compiler/src/fuzzer/input_pool.rs
```diff
@@ -0,0 +1,44 @@
...
+            return generate_input(self.num_args, &self.symbols);
+        }
+
+        let (input, _) = self
+            .inputs_and_scores
+            .choose_weighted(&mut rng, |(_, score)| *score)
+            .unwrap();
+        let mut input = input.clone();
+        mutate_input(&mut rng, &mut input, &self.symbols);
+        return input;
```
JonasWanke: `generate_input(…)` and `mutate_input(…)` could generate duplicates of the inputs we already tested. We should probably prevent that since a duplicate won't give us any new information

compiler/src/fuzzer/runner.rs
```diff
@@ -0,0 +1,133 @@
...
+            RunResult::Panicked { reason, .. } => format!("{call} panicked internally: {reason}"),
+        }
+    }
+}
+
+impl Runner {
+    pub fn new(closure_heap: &Heap, closure: Pointer, input: Input) -> Self {
+        assert!(matches!(
+            closure_heap.get(closure).data,
+            vm::Data::Closure(_)
```
JonasWanke:
```suggestion
            vm::Data::Closure(_),
```

compiler/src/fuzzer/values.rs
```diff
@@ -0,0 +1,206 @@
...
+        arguments.push(generate_value(symbols));
+    }
+    Input { arguments }
+}
+
+fn generate_value(symbols: &[String]) -> Packet {
+    let mut heap = Heap::default();
+    let address =
+        generate_value_with_complexity(&mut heap, &mut rand::thread_rng(), 100.0, symbols);
+    Packet { heap, address }
```
JonasWanke: We could join the heaps of multiple arguments, so that `Input` only has a `Heap` and a `Vec<Pointer>`

compiler/src/fuzzer/values.rs
```diff
@@ -0,0 +1,206 @@
...
+pub fn complexity_of_input(input: &Input) -> usize {
+    input
+        .arguments
+        .iter()
+        .map(|packet| complexity_of_value(&packet.heap, packet.address))
+        .sum()
+}
+fn complexity_of_value(heap: &Heap, address: Pointer) -> usize {
+    match &heap.get(address).data {
+        Data::Int(int) => int.value.magnitude().bits() as usize,
```
JonasWanke: Do we really want the $log_2(log_{10}(bits))$ or does one logarithm suffice?

compiler/src/fuzzer/values.rs
```diff
@@ -0,0 +1,206 @@
...
+                .sum::<usize>()
+                + 1
+        }
+        Data::Struct(struct_) => {
+            struct_
+                .fields
+                .iter()
+                .map(|(_, key, value)| {
+                    complexity_of_value(heap, *key) + complexity_of_value(heap, *value)
+                })
```
JonasWanke:
```suggestion
            struct_
                .iter()
                .map(|(key, value)| {
                    complexity_of_value(heap, *key) + complexity_of_value(heap, *value)
                })
```

JonasWanke (changes requested):

compiler/src/fuzzer/input.rs
```diff
@@ -0,0 +1,58 @@
...
+            let value = (*argument).hash(&heap);
+            state.write_u64(value);
+        }
+    }
+}
+impl PartialEq for Input {
+    /// This function assumes that the other input uses the same underlying
+    /// heap. This assumption should hold because all inputs generated during a
+    /// fuzzing run are saved in the same heap.
+    fn eq(&self, other: &Self) -> bool {
```
JonasWanke, line 37: https://doc.rust-lang.org/stable/std/rc/struct.Rc.html#method.ptr_eq

```suggestion
    fn eq(&self, other: &Self) -> bool {
        assert!(Rc::ptr_eq(&self.heap, &other.heap));
```

compiler/src/fuzzer/input_pool.rs
```diff
@@ -1,44 +1,62 @@
...
 impl InputPool {
     pub fn new(num_args: usize, symbols_in_heap: Vec<String>) -> Self {
+        let mut heap = Heap::default();
+
+        let mut symbols = symbols_in_heap
+            .into_iter()
+            .map(|symbol| heap.create_symbol(symbol))
+            .collect_vec();
+        if symbols.is_empty() {
+            symbols.push(heap.create_symbol("Nothing".to_string()));
```
JonasWanke:
```suggestion
            symbols.push(heap.create_nothing());
```

compiler/src/fuzzer/input_pool.rs
```diff
@@ -1,44 +1,62 @@
...
             num_args,
-            symbols: symbols_in_heap,
-            inputs_and_scores: vec![],
+            symbols,
+            input_scores: FxHashMap::default(),
         }
     }
 
     pub fn generate_new_input(&self) -> Input {
         let mut rng = ThreadRng::default();
```
JonasWanke, line 52: AFAICT, this still doesn't check whether the new input already exists in `self.input_scores`

JonasWanke (changes requested):

compiler/src/fuzzer/input_pool.rs
MarcelGarus: Oh, right 😅

JonasWanke (approved):

PR #275, opened 2023-01-22
Add helper for recursive functions

MarcelGarus:


packages/Core/controlFlow.candy
```diff
@@ -13,17 +13,22 @@ if condition then :=
   needs (function.is0 then)
   ifElse condition then {}
 
+recursive initialArg body :=
+  needs (function.is2 body) "The `body` should take the recursive function and the argument."
+  fullRec = { fullRec arg ->
+    rec = { arg -> fullRec fullRec arg }
```
JonasWanke: Maybe we should call this `recurse` to be more explicit?

packages/Core/controlFlow.candy
```diff
@@ -13,17 +13,22 @@ if condition then :=
...
 loop body :=
   needs (function.is0 body)
-  loopRec loopRec body
+  recursive Nothing { rec arg ->
+    result = function.run body
+    needs
+      ((result | equals Continue) | bool.or (result | equals Break))
+      "The `body` has to return `Continue` or `Break`."
+    if (result | equals Continue) { rec Nothing }
+  }
```
JonasWanke:
```suggestion
  recursive Nothing { recurse arg ->
    result = function.run body
    needs
      ((result | equals Continue) | bool.or (result | equals Break))
      "The `body` has to return `Continue` or `Break`."
    if (result | equals Continue) { recurse Nothing }
  }
```

packages/examples/fibonacci.candy
```diff
@@ -1,13 +1,12 @@
...
-}
 fibonacci n =
   needs (int.is n)
-  fibonacciRec fibonacciRec n
+  recursive n { rec n ->
+    ifElse (n | int.isLessThan 2) { n } {
+      rec (n | int.subtract 1)
+      | int.add (rec (n | int.subtract 2))
+    }
+  }
```
JonasWanke:
```suggestion
  recursive n { recurse n ->
    ifElse (n | int.isLessThan 2) { n } {
      recurse (n | int.subtract 1)
      | int.add (recurse (n | int.subtract 2))
    }
  }
```

JonasWanke (changes requested):

JonasWanke (approved):

PR #326, opened 2023-02-08
Find references to ints

JonasWanke:
image

This should be more useful once we support underscores and other radixes in int literals, since normal search won't work with them

### Checklist

### DIFF
--- a/compiler/language_server/src/references.rs
+++ b/compiler/language_server/src/references.rs
@@ -202,7 +202,7 @@ where
                     self.add_reference(id, DocumentHighlightKind::READ);
                 }
             },
-             Expression::Text(_) => {},
+            Expression::Text(_) => {},
             Expression::Reference(target) => {
                 if let ReferenceQuery::Id(target_id) = &self.query && target == target_id {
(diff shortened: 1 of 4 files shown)


compiler/language_server/src/references.rs
```diff
@@ -193,7 +197,12 @@ where
     }
     fn visit_expression(&mut self, id: hir::Id, expression: &Expression) {
         match expression {
-            Expression::Int(_) | Expression::Text(_) => {}
+            Expression::Int(int) =>{
+                if let ReferenceQuery::Int(_, target) = &self.query && int == target {
+                    self.add_reference(id, DocumentHighlightKind::READ);
+                }
+            },
+             Expression::Text(_) => {},
```
MarcelGarus:
```suggestion
            Expression::Text(_) => {},
```

MarcelGarus (changes requested):
What's up with the full tracer not using an address map anymore? Doesn't that make it prohibitively expensive to trace anything, like we realised in the past?

MarcelGarus (approved):

PR #328, opened 2023-02-10
Fix some bugs and make core package richer

MarcelGarus:
This PR adds some features to the core package. It also fixes bugs that were discovered while doing this and while fuzzing the compiler.


compiler/frontend/src/hir_to_mir.rs
```diff
@@ -267,27 +275,25 @@ impl<'a> LoweringContext<'a> {
...
-                    if !pattern.contains_captured_identifiers() {
-                        body.push_reference(nothing)
-                    } else {
-                        // The generated expression will be followed by at least one
-                        // identifier reference, so we don't have to generate a
-                        // reference to the result in here.
-                        self.last_pattern_result = Some((pattern_result, false));
-                        pattern_result
-                    }
+                    let pattern_result = body.push_reference(pattern_result);
```
JonasWanke: Do we actually have to push this reference? Subsequent `PatternIdentifierReference`s will refer to the correct ID using `self.ongoing_destructuring` and a `Nothing` will follow so it also can't be the surrounding body's return value

packages/Core/controlFlow.candy
```diff
@@ -32,3 +34,14 @@ loop body :=
       "The `body` has to return `Continue` or `Break`."
     if (result | equals Continue) { recurse Nothing }
   }
+
+repeat times body :=
+  needs (typeOf times | equals Int)
+  needs (times | ✨.intCompareTo 0 | equals Less | bool.not)
+  needs (function.is0 body)
```
JonasWanke, line 41: In Kotlin, the `body` accepts the repetition index as a parameter. We could even support `function.is0` _and_ `function.is1` here, but could also say that if you need the index, you should create a range (once we have that)

packages/Core/mutable.candy
```diff
@@ -0,0 +1,43 @@
...
+        requests.receivePort | channel.receive %
+          (Get, answerPort) ->
+            needs (channel.isSendPort answerPort)
+            answerPort | channel.send value
+            recurse value
+          (Set, newValue) -> recurse newValue
+          Quit -> value
+      }
+    }
+    await finalValue
```
JonasWanke: Couldn't we remove the `async`/`await` around here and call `recursive` directly?

JonasWanke (changes requested):

packages/Core/controlFlow.candy
MarcelGarus: I like the option of explicitly specifying a range.

compiler/frontend/src/hir_to_mir.rs
MarcelGarus: I simplified the code a bit by just returning a nothing. We could further simplify the HIR by just not pushing anything at all for this expression, but it makes the rest of this function more complicated (we'd need the expression compilation to produce an `Option<Id>` everywhere).

Considering that tree shaking is a thing, I'm fine with the current solution.

compiler/frontend/src/hir_to_mir.rs
```diff
@@ -268,27 +276,24 @@ impl<'a> LoweringContext<'a> {
...
-                        // identifier reference, so we don't have to generate a
-                        // reference to the result in here.
-                        self.last_pattern_result = Some((pattern_result, false));
-                        pattern_result
-                    }
+                    self.ongoing_destructuring = Some(OngoingDestructuring {
+                        result: pattern_result,
+                        is_trivial: false,
+                    });
+                    body.push_nothing()
```
JonasWanke: By moving the `self.ongoing_destructuring = …` above `body.push_if_else(…)`, we can get rid of the `body.push_nothing()`

JonasWanke (changes requested):

JonasWanke (approved):

PR #329, opened 2023-02-10
Core fixes and iterable additions

JonasWanke:
### Checklist


packages/Core/iterable.candy
```diff
@@ -31,10 +32,10 @@ isEmpty iterable :=
 generateWithStateHelper = { generateWithStateHelper state mapper ->
   {
     mapper state | result.map { result ->
-      [newItem, newState] = result
+      [item, state] = result
```
MarcelGarus: Just noticing that we could also inline the entire function using the `recursive` function.

packages/Core/text.candy
```diff
@@ -2,7 +2,7 @@ bool = use "..bool"
 [ifElse] = use "..controlFlow"
 [equals] = use "..equality"
 int = use "..int"
-iterable = use "..iterable"
+# iterable = use "..iterable"
```
MarcelGarus, line 5: Why that?

MarcelGarus (changes requested):

packages/Core/text.candy
JonasWanke: `iterable` imports `text`, so this causes a cycle. It was only used in the parameter validation of `fromUtf8`, which is currently commented out for performance reasons

packages/Core/iterable.candy
JonasWanke: We could also use that for the remaining helper functions in this file, but I'd wait for better pattern support (#295) before doing so

MarcelGarus (approved):

PR #346, opened 2023-02-23
All hail the HIR ID! Long live the HIR ID!

MarcelGarus:
This PR changes the compiler so that the HIR IDs are carried into the MIR. MIR lambdas now store the HIR ID that they originate from.

This makes debugging much easier as we know the HIR IDs of the lambdas:

image


compiler/frontend/src/mir/expression.rs
```diff
@@ -27,6 +27,7 @@ pub enum Expression {
     /// is fuzzable or not, this parameter may be used to dynamically determine
     /// who's at fault if some `needs` is not fulfilled.
     Lambda {
+        original_hirs: Vec<hir::Id>,
```
JonasWanke:
```suggestion
        original_hirs: FxHashSet<hir::Id>,
```

Makes more sense semantically, but probably not necessary

compiler/frontend/src/mir_optimize/common_subtree_elimination.rs
```diff
@@ -45,12 +47,26 @@ impl Mir {
...
                     }
                     _ => {
                         existing_entry.insert_entry(id);
                     }
                 }
             });
+
+        self.body.visit(&mut |id, expression, _| {
+            if let Expression::Lambda { original_hirs, .. } = expression && let Some(additional_hirs) = additional_lambda_hirs.remove(&id) {
+                original_hirs.extend(additional_hirs);
```
JonasWanke, line 90: I think we also have to update them for nested lambdas

JonasWanke (changes requested):

JonasWanke (approved):

PR #347, opened 2023-02-24
More optimizations

MarcelGarus:
When importing other files, we often destructure things into patterns (such as `[equals] = use "..equality"`). This wasn't properly optimized in the MIR before – the restructuring uses `builtinStructHasKey`, which wasn't constant-folded yet. This caused loooots of code to be created.

This PR fixes that by adding more constant folding, resulting in smaller code. I haven't measured the compile times yet, but it should also have a positive effect on them (all the optimizations work on smaller MIRs).

## Number of expressions in optimized MIR

when                  | type | todo | Core  | fibonacci
----------------------|------|------|-------|----------
before                |   74 | 4696 | 37303 |     37446
more constant folding |   74 | 4650 | 43690 |     43832
`StructHasKey`        |   17 |   72 |  4933 |       245
correct `typeOf`      |   47 |   72 |  4933 |       245


packages/Core/type.candy
```diff
@@ -2,4 +2,6 @@
 
 typeOf value := ✨.typeOf value
 
-is value type := equals (typeOf value) type
+is value type :=
+  needs ((typeOf type) | equals Symbol)
```
JonasWanke:
```suggestion
  needs (typeOf type | equals Symbol)
```

compiler/frontend/src/mir_optimize/constant_folding.rs
```diff
@@ -70,226 +71,285 @@ impl Mir {
...
+            let value = fields
+                .iter()
+                .rev()
+                .find(|(k, _)| k.semantically_equals(*key, visible).unwrap_or(false))
+                .map(|(_, value)| *value);
+            if let Some(value) = value {
+                Expression::Reference(value)
+            } else {
+                return Some(Err(format!(
+                    "Struct access will panic because key {:?} isn't in there.",
```
JonasWanke: Did you mean to use `Display` to get the `RichIr`'s text?

```suggestion
                    "Struct access will panic because key {} isn't in there.",
```

compiler/frontend/src/mir_optimize/constant_folding.rs
```diff
@@ -70,226 +71,285 @@ impl Mir {
...
-                    return Some(Err(format!(
-                        "Struct access will panic because key {} isn't in there.",
-                        visible.get(key_id).to_rich_ir().text,
-                    )));
+            let mut is_contained = Some(false);
+            for (k, _) in fields.iter() {
+                match k.semantically_equals(*key, visible) {
+                    Some(is_equal) => {
+                        if is_equal {
+                            is_contained = Some(true);
```
JonasWanke, line 237:
```suggestion
                            is_contained = Some(true);
                            break;
```

JonasWanke (changes requested):

JonasWanke (approved):

PR #361, opened 2023-03-06
Unify CST and RCST

JonasWanke:
### Checklist


compiler/frontend/src/cst/error.rs
```diff
@@ -0,0 +1,29 @@
...
+    IntContainsNonDigits,
+    ListItemMissesValue,
+    ListNotClosed,
+    MatchMissesCases,
+    MatchCaseMissesArrow,
+    MatchCaseMissesBody,
+    OpeningParenthesisWithoutExpression,
+    OrPatternMissesRight,
+    ParenthesisNotClosed,
+    PipeMissesCall,
```
MarcelGarus: I know you only moved the code here, but we should probably rename these as the CST doesn't know the semantic of bars.

compiler/frontend/src/cst/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    fmt::{self, Display, Formatter},
+    ops::Range,
+};
+
+mod error;
+mod id;
+mod is_multiline;
+mod kind;
+mod tree_with_ids;
+mod unwrap_whitespace_and_comment;
```
MarcelGarus, line 18: We usually sorted the `mod` parts before the `use`s. But if this is auto-inserted like this, I won't complain.

compiler/frontend/src/cst/error.rs
MarcelGarus: Oh, also some errors use `Without`, some use `Missing`. But we can fix that in another PR

compiler/language_server/src/features_candy/find_definition.rs
```diff
@@ -20,7 +20,7 @@ where
         _ => return None,
     }
 
-    let origin_hir_id = db.cst_to_hir_id(module.clone(), origin_cst.id)?;
+    let origin_hir_id = db.cst_to_hir_id(module.clone(), origin_cst.data.id)?;
```
MarcelGarus, line 23: If we also implement `Deref` for the CST, can this be removed?

MarcelGarus (changes requested):
I didn't look at `string_to_rcst` in too much detail, but it should be fine.

What also came to my mind: If we implement a `children` method on the CST that returns an `impl Iterator<Item =&Cst>`, some of the utility methods can be implemented in a much simpler way. For example `is_multiline`:

```
fn is_multiline(&self) -> bool {
  match self.kind {
    CstKind::Newline(_) => true,
    other => other.children.any(|child| child.is_multiline()),
  }
}
```

Same goes for display, finding IDs, etc.

compiler/frontend/src/cst/mod.rs
JonasWanke: When I manually touched them, I put `mod`ule declarations _after_ `use`s^^ The only official statement I can find about this topic is:

> We recommend that imports come before module declarations; […]
>
>https://github.com/rust-lang/style-team/blob/02f3c00c06c6609e3e0add3f2f15f448e12b709a/guide/items.md?plain=1#L6-L8

I'm not sure in which order they get inserted automatically, and am a bit disappointed that rustfmt doesn't reorder them. Relevant issues: https://github.com/rust-lang/rustfmt/issues/4741, https://github.com/rust-lang/rustfmt/issues/3959

compiler/language_server/src/features_candy/find_definition.rs
JonasWanke: We do implement `Derive` for `Cst` (using [<kbd>derive_more</kbd>](https://crates.io/crates/derive_more)), but `.data` is still necessary here. I'm assuming that `Deref`-ing only works for method calls and not field accesses

compiler/frontend/src/cst/error.rs
JonasWanke: Do you have a preference for either of the names?

JonasWanke:
Unfortunately, `pub fn children(&self) -> impl Iterator<Item = &Cst<D>>` is not that easy since, despite the `impl …`, all `match` arms still have to return the same iterator type. Hence, I'd move that to a separate PR (if it's even viable).

Also, I don't really understand why the formatter is failing for `candy_frontend` where it's complaining about `candy_formatter`

compiler/frontend/src/cst/error.rs
MarcelGarus: I think I prefer "misses" as it more clearly communicates there's an error

MarcelGarus: I think I prefer "misses" as it more clearly communicates there's an error

MarcelGarus (approved):

JonasWanke:
Looks like at least some cargo commands in our CI were checking the whole workspace when we executed them for a single crate. I now unified invocations per crate into one invocation each to avoid this. This also reduces our CI time

MarcelGarus (approved):

PR #371, opened 2023-03-20
Formatter

JonasWanke:
Changes in this PR:

- performance improvements
- show hints with multiline content as a single line
- code cleanup
- parser fixes
- oh, and we now have a formatter

This is only a draft PR since I still have to add support for trailing sandwich-likes. But I wanted to give you a chance to catch up since this PR is leaning slightly towards the larger end.

After that, we only have to add support for formatting text (once we update the parser to the new text formatting) and integrate it in our language server and CLI.

I switched from producing a new CST and diffing that against the old one to computing and collecting text edits directly (in `TextEdits`) since it seemed simpler. (I don't have the full comparison since I didn't implement diffing, but at least the formatting code in `ExistingWhitespace` got simpler.) I also added a bit of documentation and there are many tests for the formatter (`src/format.rs`) as well as some for `ExistingWhitespace`.

### Checklist


compiler/formatter/src/width.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    pub fn with_indent(self) -> Self {
+        Self(self.0 + 1)
+    }
+}
+impl Display for Indentation {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        for _ in 0..self.0 {
+            write!(f, "  ")?;
+        }
```
MarcelGarus:
```suggestion
        write!(f, "{}", "  ".repeat(self.0))?;
```

compiler/formatter/src/format.rs
```diff
@@ -0,0 +1,1704 @@
...
+                !matches!(
+                    cst.kind,
+                    (CstKind::Whitespace(_)
+                        | CstKind::Error {
+                            error: CstError::TooMuchWhitespace,
+                            ..
+                        }
+                        | CstKind::Newline(_)
+                        | CstKind::Comment { .. }),
+                )
```
MarcelGarus: Sure about the lint? Why can't the parens around the pattern be removed?

compiler/formatter/src/text_edits.rs
```diff
@@ -0,0 +1,126 @@
...
+        self.edits
+            .binary_search_by_key(&offset, |it| it.range.start)
+            .map(|_| true) // An edit starts at this position.
+            .unwrap_or_else(|index| {
+                self.edits
+                    .get(index)
+                    // An edit contains this position.
+                    .map(|it| it.range.contains(&offset))
+                    .unwrap_or_default()
+            })
```
MarcelGarus, line 51: Does this work for multiple inserts followed by a replacement edit containing the position? In particular, does `binary_search_by_key` always return the last entry in case there's a run of equal ones?

compiler/formatter/src/existing_parentheses.rs
```diff
@@ -0,0 +1,244 @@
...
+                                opening: merge(
+                                    edits,
+                                    old_opening_parenthesis,
+                                    new_opening_parenthesis,
+                                ),
+                                closing: merge(
+                                    edits,
+                                    new_closing_parenthesis,
+                                    old_closing_parenthesis,
+                                ),
```
MarcelGarus, line 101: Is it correct that new and old are switched in the closing case?

compiler/formatter/src/existing_whitespace.rs
```diff
@@ -0,0 +1,568 @@
...
+    }
+    pub fn into_empty_and_move_comments_to(
+        self,
+        edits: &mut TextEdits,
+        other: &mut ExistingWhitespace<'a>,
+    ) {
+        if self.is_empty()
+        {
+            return;
+        }
```
MarcelGarus: Weirdly formatted

compiler/formatter/src/existing_whitespace.rs
```diff
@@ -0,0 +1,568 @@
...
+                .last()
+                .map(|it| it.data.span.end)
+                .unwrap_or_else(|| other.start_offset);
+            if self.start_offset == other_end_offset
+                && other.adopted_whitespace_after.is_empty()
+                && self.adopted_whitespace_before.is_empty()
+                && self.adopted_whitespace_after.is_empty()
+                && !edits.has_edit_at(self.start_offset)
+            {
+                // Simple case: The whitespace is adopted by directly precedinb whitespace.
```
MarcelGarus:
```suggestion
                // Simple case: The whitespace is adopted by directly preceding whitespace.
```

compiler/formatter/src/format.rs
```diff
@@ -0,0 +1,1704 @@
...
+                &info.with_indent().with_indent(),
+            );
+            MaybeSandwichLikeArgument::Other {
+                argument,
+                min_singleline_width: Width::multiline(None),
+            }
+        } else if is_last
+            && matches!(
+                argument.kind,
+                (CstKind::List { .. } | CstKind::Struct { .. } | CstKind::Lambda { .. }),
```
MarcelGarus: Sure the parentheses can't be removed here?

compiler/formatter/src/format.rs
```diff
@@ -0,0 +1,1704 @@
...
+            | CstKind::Whitespace(_)
+            | CstKind::Newline(_)
+            | CstKind::Comment { .. } => None,
+            CstKind::TrailingWhitespace { child, .. } => child.precedence(),
+            CstKind::Identifier(_) | CstKind::Symbol(_) | CstKind::Int { .. } => {
+                Some(PrecedenceCategory::High)
+            }
+            CstKind::OpeningText { .. } | CstKind::ClosingText { .. } => None,
+            CstKind::Text { .. } => Some(PrecedenceCategory::High),
+            CstKind::TextPart(_) => todo!(),
```
MarcelGarus: .

compiler/formatter/src/format.rs
```diff
@@ -0,0 +1,1704 @@
...
+        test("foo | bar", "foo | bar\n");
+        test("foo|bar", "foo | bar\n");
+        test("foo  |  bar", "foo | bar\n");
+        test("foo\n\n|   bar", "foo | bar\n");
+        test("foo | (bar)", "foo | bar\n");
+        test("foo | (\n  bar\n)", "foo | bar\n");
+        test("foo | (bar baz)", "foo | (bar baz)\n");
+        test("foo | (bar | baz)", "foo | (bar | baz)\n");
+        test(
+            "veryVeryVeryVeryVeryVeryVeryVeryLongReceiver | (veryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongFunction)",
```
MarcelGarus: For easier readability, I'd rename these to `loooooooooooooooooooooooooongReceiver` and similar

MarcelGarus (changes requested):

compiler/formatter/src/text_edits.rs
JonasWanke: There's never a run of entries with the same start offset, since they get merged: `change(…)`  checks if the new edit overlaps or touches any existing edit, in which case the code either panics (for overlapping, incompatible changes) or merges them into a single edit

compiler/formatter/src/existing_parentheses.rs
JonasWanke: Yes: The order is the same as they appear in the source code so that we prefer to keep the left-most parenthesis and move all comments there if necessary

compiler/formatter/src/format.rs
JonasWanke: There's a bug in the newer nightly versions of Rust that appears to no longer parse an or-pattern in a macro:

image

This surfaces in multiple places for us

MarcelGarus (approved):

PR #390, opened 2023-04-16
Support packages

MarcelGarus:
This adds support for package markers and managed packages.

By having a `_package.candy` file in a Candy module (aka, having a folder with both a `_.candy` and a `_package.candy` file), this module is turned into a package. Modules from inside the package can't relatively import anything outside of the module.

Also, there now exists a package folder (which is just hardcoded to our `packages` folder relative to the executable). If you just `use "Something"`, then the `Something` folder is included.

This PR also makes the CLI easier to use. In particular, you can now just `cargo run` anywhere in our repo without any additional options to build and run the CLI. If you execute `cargo run -- run` without additional parameters, the package surrounding the current working directory is run.


compiler/cli/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "candy_cli"
 version = "0.1.0"
 edition = "2021"
 rust-version = "1.56"
+default-run = "candy"
```
JonasWanke, line 6: Is this really required when there's only a single binary?

compiler/cli/src/main.rs
```diff
@@ -67,9 +70,14 @@ struct CandyBuildOptions {
     tracing: bool,
 
     #[arg(value_hint = ValueHint::FilePath)]
-    file: PathBuf,
+    path: Option<PathBuf>,
 }
 
+/// Run a Candy program.
+///
+/// This command runs the given file, or if no file is provided, the package of
```
JonasWanke:
```suggestion
/// This command runs the given file, or, if no file is provided, the package of
```

compiler/cli/src/main.rs
```diff
@@ -78,17 +86,26 @@ struct CandyRunOptions {
     #[arg(long)]
     tracing: bool,
 
+    /// The file or package to run. If non is provided, the package of your
```
JonasWanke:
```suggestion
    /// The file or package to run. If none is provided, the package of your
```

compiler/frontend/src/module/module.rs
```diff
@@ -17,28 +21,38 @@ pub enum ModuleKind {
 }
 
 impl Module {
-    pub fn from_package_root_and_file(
-        package_root: PathBuf,
-        file: PathBuf,
+    pub fn from_package_name(name: &str) -> Self {
+        Module {
+            package: Package::Managed(name.into()),
```
JonasWanke: If we're going to need an owned version of the name, we can accept that directly:

```suggestion
    pub fn from_package_name(name: String) -> Self {
        Module {
            package: Package::Managed(name),
```

compiler/frontend/src/module/use_path.rs
```diff
@@ -1,56 +1,84 @@
...
     }
 }
+impl Display for UsePath {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            UsePath::Managed(name) => write!(f, "{}", name),
+            UsePath::Relative {
+                parent_navigations,
+                path,
+            } => write!(f, "{}{}", ".".repeat(1 + parent_navigations), path),
```
JonasWanke:
```suggestion
            } => write!(f, "{}{path}", ".".repeat(1 + parent_navigations)),
```

compiler/language_server/src/features_ir.rs
```diff
@@ -55,13 +49,8 @@ pub struct ViewIrParams {
...
 impl Server {
     pub async fn candy_view_ir(&self, params: ViewIrParams) -> jsonrpc::Result<String> {
-        let project_directory = {
-            let state = self.state.read().await;
-            state.require_running().project_directory.to_owned()
-        };
-        let config = IrConfig::decode(&params.uri, project_directory);
-
         let state = self.state.read().await;
+        let config = IrConfig::decode(&params.uri, &state.require_running().packages_path);
```
JonasWanke, line 53:
```suggestion
        let config = IrConfig::decode(&params.uri, &state.require_running().packages_path);
        let state = self.state.read().await;
```

(Swap the lines to keep the lock for a shorter time)

compiler/language_server/src/server.rs
```diff
@@ -183,32 +195,31 @@ impl LanguageServer for Server {
...
 
+        let packages_path = {
+            let options = params
+                .initialization_options
+                .as_ref()
+                .expect("No initialization options provided.")
+                .as_object()
+                .unwrap();
+            let path_string = options.get("packagesPath").unwrap().as_str().unwrap();
+            PathBuf::from_str(path_string).expect("Invalid path")
```
JonasWanke:
```suggestion
            options.get("packagesPath").unwrap().as_str().unwrap().into()
```

JonasWanke (changes requested):

JonasWanke (approved):

PR #394, opened 2023-04-17
Add Tags

### LINKED ISSUE #282: Tags
Instead of symbols, we want to have tags that can also have an optional associated value. Our current symbols will just be named tags. Tags without an associated value can be called just like functions to associate a value with them.

```candy
tagWithoutValue = Foo
tagWithValue = Foo 3
newResult = Ok 4
error = Error "That didn't work."
dynamicallyCreatedTag = tagWithoutValue 5
nestedTag = Foo (Bar 3)
```

skyz1:
Closes: #282 

This PR allows symbols to be associated with a value by calling them with that value.
image

### Checklist


compiler/frontend/src/builtin_functions.rs
```diff
@@ -36,6 +36,9 @@ pub enum BuiltinFunction {
     StructGet,           // struct key -> value
     StructGetKeys,       // struct -> listOfKeys
     StructHasKey,        // struct key -> booleanSymbol
+    TagGetSymbol,        // symbol -> symbol
```
MarcelGarus: To the user, I wouldn't even mention the word symbol and instead, use tags consistently (which either have values or don't). This builtin could then be named `tagWithoutValue` (or something like that).

I'd also rename the other references here (for example, change the comment of `structHasKey` to `struct key -> booleanTag`).

Of course, the rest of our internal compiler code would continue to use "symbol" like it already does.

compiler/frontend/src/builtin_functions.rs
```diff
@@ -36,6 +36,9 @@ pub enum BuiltinFunction {
...
+    TagHasValue,         // symbol -> booleanSymbol
+    TagGetValue,         // symbol -> any
```
JonasWanke: Please keep them ordered alphabetically

compiler/vm/src/heap/mod.rs
```diff
@@ -307,8 +308,8 @@ impl Heap {
     pub fn create_text(&mut self, text: String) -> Pointer {
         self.create(Data::Text(Text { value: text }))
     }
-    pub fn create_symbol(&mut self, symbol: String) -> Pointer {
-        self.create(Data::Symbol(Symbol { value: symbol }))
+    pub fn create_tag(&mut self, symbol: String, value: Option<Pointer>) -> Pointer {
```
JonasWanke: By slightly modifying the argument type, this function can also be called with a `Pointer` directly since any `T` `impl`s `Into<Option<T>>`:

```suggestion
    pub fn create_tag(&mut self, symbol: String, value: impl Into<Option<Pointer>>) -> Pointer {
```

compiler/vm/src/fiber.rs
```diff
@@ -527,6 +528,21 @@ impl Fiber {
                 self.heap.drop(callee);
                 self.run_builtin_function(&builtin, &arguments, responsible);
             }
+            Data::Tag(Tag { symbol, .. }) => {
+                if let [value] = arguments.as_slice() {
+                    let tag = self.heap.create_tag(symbol.to_string(), Some(*value));
+                    self.data_stack.push(tag);
+                    self.heap.dup(*value);
```
JonasWanke: I think that this should error when trying to overwrite the tag's current value. We could then point the user to use `tag.withValue` instead

packages/Core/tag.candy
```diff
@@ -0,0 +1,27 @@
...
+
+withSymbol tag symbolTag :=
+  needs (is tag)
+  needs (is symbolTag)
+  ifElse (tag | hasValue) { symbolTag (tag | getValue) } { symbolTag | getSymbol }
+
+withValue tag valueTag :=
+  needs (is tag)
+  needs (is valueTag)
+  ifElse (valueTag | hasValue) { tag (valueTag | getValue) } { tag | getSymbol }
```
MarcelGarus: I would have expected this to take a tag and a value, for example, `tag.withValue Foo 5` (= `Foo 5`) or `tag.withValue (Foo 2) 3` (= `Foo 3`).

compiler/vm/src/fiber.rs
```diff
@@ -527,6 +528,21 @@ impl Fiber {
...
+                } else {
+                    self.panic(
+                        format!(
+                            "Assigning the value of a tag expects 1 parameter, but you provided {} arguments.",
```
JonasWanke:
```suggestion
                            "A tag can only hold exactly one value, but you called it with {} arguments.",
```

MarcelGarus (changes requested):

compiler/vm/src/heap/object.rs
```diff
@@ -299,7 +316,9 @@ impl Pointer {
         match &heap.get(*self).data {
             Data::Int(int) => format!("{}", int.value),
             Data::Text(text) => format!("\"{}\"", text.value),
-            Data::Symbol(symbol) => symbol.value.to_string(),
+            Data::Tag(Tag { symbol, value }) => value.map_or(symbol.to_string(), |value| {
+                format!("{}({})", symbol, value.format_helper(heap, is_debug))
```
JonasWanke: To stay closer to the source code, I'd add a whitespace before the parentheses. We can add a smarter stringification later on that tracks precedence so that parentheses are only added where necessary

```suggestion
                format!("{} ({})", symbol, value.format_helper(heap, is_debug))
```

compiler/vm/src/lir.rs
```diff
@@ -26,7 +26,7 @@ pub enum Instruction {
     /// Pushes a text.
     CreateText(String),
 
-    /// Pushes a symbol.
+    /// Pushes a empty tag.
```
JonasWanke:
```suggestion
    /// Pushes an empty tag.
```

packages/Core/tag.candy
```diff
@@ -0,0 +1,27 @@
+type = use "..type"
+[ifElse] = use "..controlFlow"
```
JonasWanke: We order imports alphabetically by path:

```suggestion
[ifElse] = use "..controlFlow"
type = use "..type"
```

JonasWanke (changes requested):

JonasWanke (approved):

compiler/vm/src/fiber.rs
```diff
@@ -528,15 +527,23 @@ impl Fiber {
                 self.heap.drop(callee);
                 self.run_builtin_function(&builtin, &arguments, responsible);
             }
-            Data::Tag(Tag { symbol, .. }) => {
+            Data::Tag(Tag { symbol, value }) => {
+                if value.is_some() {
+                    self.panic(
+                        "A tags value cannot be overwritten by calling it. Use tag.withValue instead.".to_string(),
```
MarcelGarus:
```suggestion
                        "A tag's value cannot be overwritten by calling it. Use `tag.withValue` instead.".to_string(),
```

MarcelGarus (changes requested):

packages/Core/tag.candy
```diff
@@ -0,0 +1,28 @@
...
+withSymbol tag symbolTag :=
+  needs (is tag)
+  needs (is symbolTag)
+  ifElse (tag | hasValue) { symbolTag (tag | getValue) } { symbolTag | withoutValue }
+
+withValue tag value :=
+  needs (is tag)
+  ## TODO: Fix formatter to allow (tag | withoutValue) value
+  tagWithoutValue = tag | withoutValue
+  tagWithoutValue value
```
JonasWanke: I just fixed this in #404

```suggestion
  (tag | withoutValue) value
```

MarcelGarus (approved):

PR #412, opened 2023-04-24
Bugfixes

### LINKED ISSUE #402: Package support breaks candy on Windows
**Describe the bug**
#390 introduced a bug where the module resolution inside packages fails to detect if a file is located in the package on Windows. 

The underlying problem is the platform specific behavior of `fs::canonicalize` (see [the documentation](https://dev-doc.rust-lang.org/beta/std/fs/fn.canonicalize.html#platform-specific-behavior)). This function returns a UNC-Path prefixed with `\\?\` while `package.to_path` returns a DOS-Path without the prefix. `strip_prefix` then returns an error and the compiler thinks the file is not located in the package.

The [dunce crate](https://lib.rs/crates/dunce) was created to combat this issue, but converting the UNC-path to a URL and back to a file path seems to be a quick workaround.

```rust
let canonicalized = Url::from_file_path(canonicalized)
    .expect("Invalid file path.")
    .to_file_path()
    .expect("File path conversion error.");
```

JonasWanke:
Closes: #402

### Checklist


compiler/frontend/src/module/module.rs
```diff
@@ -161,3 +159,25 @@ impl ToRichIr for Module {
...
+            ModuleFromPathError::NotFound(path) => {
+                write!(
+                    f,
+                    "File `{}` does not exist or its path is invalid.",
+                    path.to_string_lossy(),
+                )
+            }
+            ModuleFromPathError::NotInPackage(path) => {
+                write!(f, "File {:?} is not located in the package.", path)
+            }
```
MarcelGarus: Why display paths differently for both error kinds?

MarcelGarus (approved):

PR #417, opened 2023-04-25
Support renaming identifiers

JonasWanke:
Progress towards #308

| Rename | Result |
|-|-|
| image | image |

Renaming only works for identifiers and only looks at the current file. I also thought about supporting symbol renames, but that gets more complex due to their occurrences in struct accesses.

### Checklist

### DIFF
--- a/compiler/language_server/src/server.rs
+++ b/compiler/language_server/src/server.rs
@@ -537,23 +537,17 @@ impl LanguageServer for Server {
                 params.new_name,
             )
             .await;
-        result
-            .map(|it| {
-                Some(WorkspaceEdit {
-                    changes: Some(it),
-                    ..Default::default()
-                })
-            })
-            .map_err(|it| {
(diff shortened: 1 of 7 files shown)


compiler/language_server/src/server.rs
```diff
@@ -500,15 +513,57 @@ impl LanguageServer for Server {
...
+            .map_err(|it| {
+                let message = match it {
+                    RenameError::NewNameInvalid => "The new name is not valid.",
+                };
+                jsonrpc::Error {
+                    code: jsonrpc::ErrorCode::InvalidParams,
+                    message: message.to_string(),
+                    data: None,
+                }
+            })
```
MarcelGarus: This could also just be a match

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #418, opened 2023-04-25
Replace expanduser with shellexpand

### LINKED ISSUE #402: Package support breaks candy on Windows

skyz1:
Closes: #402 

This PR replaces the `expanduser` crate added in #412 with `shellexpand`, since `expanduser` doesn't compile on windows.

However `shellexpand` doesn't expand `~user` with the home directory of `user`. If this leads to problems, conditional compilation could be used in the future to bring that feature back in.

### Checklist


compiler/frontend/src/module/package.rs
```diff
@@ -59,9 +59,7 @@ impl TryFrom<&str> for PackagesPath {
     type Error = String;
 
     fn try_from(path: &str) -> Result<Self, Self::Error> {
-        expanduser(path)
-            .map_err(|it| it.to_string())
-            .and_then(|it| it.as_path().try_into())
+        PathBuf::from(tilde(path).to_string()).as_path().try_into()
```
JonasWanke: We can avoid the allocation behind `to_string()`, the borrowed `Path` suffices:

```suggestion
        Path::new(tilde(path).as_ref()).try_into()
```

JonasWanke (changes requested):

JonasWanke (approved):

PR #430, opened 2023-05-05
Apply more suggestions

MarcelGarus:
Applies some suggestions post-merging for last PR

### DIFF
--- a/compiler/fuzzer/src/runner.rs
+++ b/compiler/fuzzer/src/runner.rs
@@ -2,7 +2,7 @@ use candy_frontend::hir::Id;
 use candy_vm::{
     self,
     context::{CombiningExecutionController, CountingExecutionController, ExecutionController},
-    heap::{Closure, HirId},
+    heap::{Closure, HirId, InlineObjectSliceCloneToHeap},
     lir::Lir,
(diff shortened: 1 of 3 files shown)


compiler/fuzzer/src/runner.rs
```diff
@@ -52,13 +52,15 @@ impl RunResult {
 impl<L: Borrow<Lir>> Runner<L> {
     pub fn new(lir: L, closure: Closure, input: Input) -> Self {
         let (mut heap, constant_mapping) = lir.borrow().constant_heap.clone();
+        let closure = closure.clone_to_heap(&mut heap).try_into().unwrap();
+        let arguments = input.arguments.clone_to_heap(&mut heap);
```
JonasWanke: These could share an address mapping. I'm even thinking about always requiring one so that we can't forget this

JonasWanke (changes requested):

compiler/fuzzer/src/runner.rs
MarcelGarus: True. I think always requiring a mapping would also be nice, but maybe in a future PR?

JonasWanke (approved):

PR #444, opened 2023-05-12
Fix fuzzer segmentation fault

MarcelGarus:
Adds some fixes so that the fuzzer command doesn't segfault anymore.
The printing of fuzzed stack traces is commented out for now.

### Checklist


compiler/fuzzer/src/lib.rs
```diff
@@ -42,11 +39,11 @@ where
...
     let lir = Arc::new(lir);
 
-    let fuzzables = {
+    let (_heap, fuzzables) = {
         let mut tracer = FuzzablesFinder::default();
-        let mut vm = Vm::for_module(lir.clone(), &mut tracer);
-        vm.run(&mut RunForever, &mut tracer);
-        tracer.into_fuzzables()
+        let result = Vm::for_module(lir.clone(), &mut tracer).run_until_completion(&mut tracer);
+        dbg!(result.reason);
```
JonasWanke:
```suggestion
```

compiler/cli/src/fuzz.rs
```diff
@@ -37,11 +37,11 @@ pub(crate) fn fuzz(options: Options) -> ProgramResult {
...
-        error!("These are the failing cases:");
-        for case in failing_cases {
-            error!("");
-            case.dump(&db);
-        }
+        // error!("These are the failing cases:");
+        // for case in failing_cases {
+        //     error!("");
+        //     case.dump(&db);
+        // }
```
JonasWanke, line 44: _Why_ is this commented out?

compiler/fuzzer/src/lib.rs
```diff
@@ -75,7 +72,7 @@ where
                     panic,
                     tracer,
                 };
-                case.dump(db);
+                // case.dump(db);
```
JonasWanke, line 74: Why is this commented out?

compiler/fuzzer/src/values.rs
```diff
@@ -61,7 +61,8 @@ fn generate_value_with_complexity(
         }
         6 => builtin_functions::VALUES[rng.gen_range(0..builtin_functions::VALUES.len())].into(),
         _ => unreachable!(),
-    }
+    };
+    value
```
JonasWanke: This change can probably be reverted?

JonasWanke (changes requested):

compiler/cli/src/fuzz.rs
MarcelGarus: Printing the stack trace currently segfaults for some reason. That also happens on main, so this PR is still an improvement, but I commented the segfaulting code out so that fuzzing at least works (it does report the function and failing inputs).
If you want to, I can also uncomment this so we don't forget to fix the segfault.

JonasWanke: Can you please add a TODO or an issue for tracking this?

MarcelGarus: Done: https://github.com/candy-lang/candy/issues/458

JonasWanke (approved):

PR #463, opened 2023-05-22
Add coverage-guided fuzzing

MarcelGarus:
Fuzzing now tracks the coverage and improvements of the coverage, not the number of executed instructions. Essentially, inputs that increase coverage are favored. This is analogous to how AFL works.

### Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

PR #464, opened 2023-05-22
Support match of tag with value

### LINKED ISSUE #396: Support tags in MIR
Allow constant evaluation to create tags in the MIR

### LINKED ISSUE #469: Tag matching with values

MarcelGarus:
Depends on #463

Closes: #469, closes #396

In a match, we only supported simple tags, not tags with values. Now, match patterns can also include tags with values.

### Checklist


packages/Core/result.candy
```diff
@@ -3,47 +3,47 @@ bool = use "..bool"
...
-is value := struct.is value | bool.lazyAnd {
-  value | struct.hasKey Type | bool.and (value | struct.hasKey Value) | bool.lazyAnd {
-    type = value.type
-    equals type Ok | bool.or (equals type Error)
-  }
-}
+is value :=
+  value %
+    Ok _ -> True
+    Error _ -> True
```
JonasWanke:
```suggestion
    Ok _ | Error _ -> True
```

compiler/frontend/src/hir_to_mir.rs
```diff
@@ -555,9 +555,50 @@ impl<'a> PatternLoweringContext<'a> {
                 let expected = body.push_text(text.to_owned());
                 self.compile_exact_value(body, expression, expected)
             }
-            hir::Pattern::Symbol(symbol) => {
-                let expected = body.push_symbol(symbol.to_owned());
-                self.compile_exact_value(body, expression, expected)
+            hir::Pattern::Tag { symbol, value } => {
+                self.compile_verify_type_condition(body, expression, "Tag".to_string(), |body| {
+                    let builtin_tag_without_value =
+                    body.push_builtin(BuiltinFunction::TagWithoutValue);
```
JonasWanke: Can you please run `rustfmt` on this file in the console to see why it doesn't want to format it?

compiler/frontend/src/hir_to_mir.rs
```diff
@@ -555,9 +555,50 @@ impl<'a> PatternLoweringContext<'a> {
...
+                                self.compile(body, actual_value, value);
+                            }
+                        }, |body, _, _| {
+                            if value.is_some() {
+                                vec![
+                                    body.push_text("Expected tag to have a value, but it doesn't have any.".to_string()),
+                                    ]
+                                } else {
+                                    vec![
+                                    body.push_text("Expected tag to not have a value, but it has one.".to_string()),
```
JonasWanke: Maybe show the value as well?

compiler/frontend/src/hir_to_mir.rs
```diff
@@ -824,7 +865,9 @@ impl<'a> PatternLoweringContext<'a> {
             }
             hir::Pattern::Int(int) => body.push_int(int.to_owned().into()),
             hir::Pattern::Text(text) => body.push_text(text.to_owned()),
-            hir::Pattern::Symbol(symbol) => body.push_symbol(symbol.to_owned()),
+            hir::Pattern::Tag { .. } => {
+                panic!("Tags can't be used in this part of a pattern.")
+            }
```
JonasWanke: Shouldn't they? E.g., `[Foo: bar]` would try to compile the tag `Foo` as a key expression

compiler/frontend/src/mir/expression.rs
```diff
@@ -213,9 +225,14 @@ impl ToRichIr for Expression {
                     builder.push(format!(r#""{}""#, text), TokenType::Text, EnumSet::empty());
                 builder.push_reference(text.to_owned(), range);
             }
-            Expression::Symbol(symbol) => {
-                let range = builder.push(symbol, TokenType::Symbol, EnumSet::empty());
+            Expression::Tag { symbol, value } => {
+                let range = builder.push(format!("{symbol}"), TokenType::Symbol, EnumSet::empty());
                 builder.push_reference(ReferenceKey::Symbol(symbol.to_owned()), range);
+                builder.push(" ", None, EnumSet::empty());
```
JonasWanke: Looks like this one is too much

compiler/frontend/src/mir_optimize/cleanup.rs
```diff
@@ -62,7 +62,6 @@ impl Mir {
                 (Expression::Builtin(a), Expression::Builtin(b)) => {
                     format!("{a:?}").cmp(&format!("{b:?}"))
                 }
-                (Expression::Symbol(a), Expression::Symbol(b)) => a.cmp(b),
```
JonasWanke, line 65: Something like this should work: `a.symbol.cmp(b.symbol).then_with(|| a.value.cmp(b.value))`

compiler/frontend/src/mir_optimize/constant_folding.rs
```diff
@@ -70,6 +76,11 @@ pub fn fold_constants(
     *expression = evaluated_call;
 }
 
+enum BuiltinResult {
+    Returns(Expression),
+    Panics(String),
+}
```
JonasWanke, line 82: We could also `impl From<Expression> for Option<BuiltinResult>` to make `run_builtin(…)` more concise

compiler/frontend/src/mir_optimize/constant_folding.rs
```diff
@@ -159,9 +172,9 @@ fn run_builtin(
             };
 
             let Some(value) = index.to_usize().and_then(|index| list.get(index)) else {
-                return Some(Err(format!("List access will panic because index {} is out of bounds.", index)));
+                return Some(Panics(format!("List access will panic because index {} is out of bounds.", index)));
```
JonasWanke:
```suggestion
                return Some(Panics(format!("List access will panic because index {index} is out of bounds.")));
```

compiler/vm/src/mir_to_lir.rs
```diff
@@ -157,9 +157,24 @@ impl<'c> LoweringContext<'c> {
...
+                        } else {
+                            self.emit_reference_to(*value);
+                            self.emit(id, Instruction::CreateTag { symbol });
+                        }
+                    }
+                    None => {
+                        let tag = Tag::create(&mut self.lir.constant_heap, symbol, None);
+                        self.constants.insert(id, tag.into());
+                    }
+                }
```
JonasWanke: Maybe the outer `match` on an `Option` could also be converted to an `if let` for consistency

JonasWanke (changes requested):

compiler/frontend/src/hir_to_mir.rs
MarcelGarus: Running `rustfmt` on the console also doesn't formats this (and prints no errors or warnings, even with `-v`). I just formatted this manually.

compiler/frontend/src/mir_optimize/cleanup.rs
MarcelGarus: For now, tags with values are not constant-lifted and not reordered because they may reference other expressions and thereby get invalid by reordering, even if they're constant.
Only tags without values are moved. I fixed it so that they are also sorted according to the symbol.

compiler/frontend/src/mir_optimize/constant_folding.rs
MarcelGarus: If I understand your plan correctly, then we'd have lots of `expr.into()` instead of `Returns(expr)` in the `run_builtin`? In that case, I prefer the explicit `Returns` one.

compiler/frontend/src/hir_to_mir.rs
```diff
@@ -581,10 +581,15 @@ impl<'a> PatternLoweringContext<'a> {
...
-                                    body.push_text("Expected tag to not have a value, but it has one.".to_string()),
+                                ]
+                            } else {
+                                let builtin_tag_get_value = body.push_builtin(BuiltinFunction::TagGetValue);
+                                let actual_value = body.push_call(builtin_tag_get_value, vec![expression], self.responsible);
+                                let builtin_to_debug_text = body.push_builtin(BuiltinFunction::ToDebugText);
+                                let actual_value_text = body.push_call(builtin_to_debug_text, vec![actual_value], self.responsible);
+                                vec![
+                                    body.push_text("Expected tag to not have a value, but it has one: ".to_string()),
+                                    actual_value_text,
```
JonasWanke:
```suggestion
                                    body.push_text("Expected tag to not have a value, but it has one: `".to_string()),
                                    actual_value_text,
                                    body.push_text("`.".to_string()),
```

compiler/frontend/src/mir_optimize/constant_folding.rs
JonasWanke: `Returns(expr).into()` would become `expr.into()`, but we can also leave this as is

JonasWanke (changes requested):

JonasWanke (approved):

PR #488, opened 2023-06-04
Fix pipe evaluation order

### LINKED ISSUE #292: Fix the evaluation order of pipe expression by keeping it in the AST
Currently, `foo 0 | (bar 1)` first evaluates `bar 1`, then `foo 0`, and then calls the result of `bar 1` with the result of `foo 0`. This is because pipes are not preserved in the AST, instead the expression gets lowered to the equivalent of `(bar 1) (foo 0)`.

We should keep pipe expressions in the AST and then fix the AST to HIR lowering so that the above code evaluates `foo 0`, then `bar 1`, and then calls the result of `bar 1` with the result of `foo 0`.

skyz1:
Closes: #292 

This PR fixes some problems of the pipe expression:
1. The value of the left side of the pipe expression now only gets prepended to the argument list of the call if the right side is a function call with low precedence, i.e. `1 | foo 2` is the same as `foo 1 2` and `1 | (foo 2)` is the same as `(foo 2) 1`.
2. The sub-expressions now get evaluated from left to right, i.e. `foo 0 | (bar 1)` gets evaluated in the order of `foo 0`, then `bar 1`, and finally `bar 1` gets called with the result of `foo 0`

### Checklist


compiler/frontend/src/ast.rs
```diff
@@ -110,6 +110,7 @@ pub struct Function {
 pub struct Call {
     pub receiver: Box<Ast>,
     pub arguments: Vec<Ast>,
+    pub pipe: bool,
```
JonasWanke: When reading the name elsewhere in code, this makes it clearer that it doesn't store a representation of the pipe, but only whether a pipe was used or not. At least for Dart, there exists an official recommendation: [PREFER starting a boolean variable or property comment with “Whether” followed by a noun or gerund phrase](https://dart.dev/effective-dart/documentation#prefer-starting-a-boolean-variable-or-property-comment-with-whether-followed-by-a-noun-or-gerund-phrase)

```suggestion
    pub is_pipe: bool,
```

JonasWanke (changes requested):

JonasWanke (approved):

PR #578, opened 2023-07-20
LLVM-based Compiler

### LINKED ISSUE #298: Support LLVM

ctiedt:
Closes: #298 

This PR introduces an initial LLVM-based compiler for Candy using the inkwell crate. While the compiler does not support all instructions yet, simple examples are already runnable.


compiler/backend_inkwell/README.md
```diff
@@ -0,0 +1,9 @@
+# Candy Inkwell Compiler
+
+This crate provides an LLVM-based backend for the Candy compiler
+using the inkwell crate. It is highly experimental and still
+doesn't support many of Candy's features.
+
+## Requirements
+
+You currently need to have LLVM 15 installed on your system. You also need to have the `llc` and `clang` commands available.
```
JonasWanke: In Markdown, we use line breaks after each sentence:

```suggestion
This crate provides an LLVM-based backend for the Candy compiler using the inkwell crate.
It is highly experimental and still doesn't support many of Candy's features.

## Requirements

You currently need to have LLVM 15 installed on your system.
You also need to have the `llc` and `clang` commands available.
```

compiler/backend_inkwell/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "backend_inkwell"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
JonasWanke:
```suggestion
```

compiler/backend_inkwell/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+[dependencies]
+candy_frontend = { version = "0.1.0", path = "../frontend" }
+inkwell = { version = "0.2.0", features = ["llvm15-0"] }
+llvm-sys = { version = "*", features = ["prefer-dynamic"] }
```
JonasWanke: Trailing newline

```suggestion
llvm-sys = { version = "*", features = ["prefer-dynamic"] }

```

compiler/backend_inkwell/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "backend_inkwell"
```
JonasWanke: Our packages start with `candy_`:

```suggestion
name = "candy_backend_inkwell"
```

compiler/backend_inkwell/Cargo.toml
JonasWanke: Version 150 corresponds to LLVM 15.0.x

```suggestion
llvm-sys = { version = "150", features = ["prefer-dynamic"] }
```

compiler/backend_inkwell/candy_rt/Makefile
```diff
@@ -0,0 +1,10 @@
+CC=clang
+
+candy_rt.a: candy_rt.o candy_builtin.o
+	ar rcs $@ $^
+
+%.o: %.c %.h
+
+.PHONY: clean
+clean:
+	rm -f *.o *.a
```
JonasWanke: Trailing newline

```suggestion
	rm -f *.o *.a

```

compiler/backend_inkwell/candy_rt/candy_rt.c
```diff
@@ -0,0 +1,204 @@
...
+        if (value->type == CANDY_TYPE_TAG || value->type == CANDY_TYPE_TEXT)
+        {
+            free(value->value.text);
+        }
+        // List and struct entries may not be freed as part of freeing
+        // the list/struct, because they will be freed on their own
+        // at the end of the main function.
+        free(value);
+    }
+}
```
JonasWanke: Please add trailing newlines

compiler/backend_inkwell/candy_rt/candy_rt.h
```diff
@@ -0,0 +1,72 @@
...
+typedef struct
+{
+    void *environment;
+    struct candy_value *(*function)(struct candy_value *);
+} candy_function_t;
+
+typedef struct
+{
+    struct candy_value **keys;
+    struct candy_value **values;
```
JonasWanke: Would this also work and get rid of `candy_value`?

```suggestion
    candy_value_t **keys;
    candy_value_t **values;
```

.gitignore
```diff
@@ -44,3 +44,9 @@ coverage/
 .buildlog/
 .history
 .svn/
+
+# LLVM Artefacts
+*.o
+*.a
+*.ll
+*.s
```
JonasWanke:
```suggestion
*.s

```

compiler/backend_inkwell/candy_rt/candy_builtin.c
```diff
@@ -0,0 +1,157 @@
...
+        candy_function then_function = (then->value).function.function;
+        candy_value_t *environment = (then->value).function.environment;
+        return then_function(environment);
+    }
+    else
+    {
+        candy_function otherwise_function = (otherwise->value).function.function;
+        candy_value_t *environment = (otherwise->value).function.environment;
+        return otherwise_function(environment);
+    }
```
JonasWanke:
```suggestion
    candy_value_t *body = candy_tag_to_bool(condition) ? then : otherwise;
    candy_function function = (body->value).function.function;
    candy_value_t *environment = (body->value).function.environment;
    return function(environment);
```

compiler/backend_inkwell/candy_rt/candy_builtin.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+candy_value_t *candy_builtin_int_subtract(candy_value_t *left, candy_value_t *right)
+{
+    return make_candy_int(left->value.integer - right->value.integer);
+}
+
+candy_value_t *candy_builtin_int_bit_length(candy_value_t *value)
+{
+    return make_candy_int(64);
```
JonasWanke: `Builtins/_.candy`'s doc comment: “Determines the fewest bits necessary to express this integer, not including the sign.” So, for `12`, it should return `4` because the binary representation needs at least four bits: $12_{10} = 1100_2$

compiler/backend_inkwell/candy_rt/candy_builtin.c
```diff
@@ -0,0 +1,157 @@
...
+
+candy_value_t *candy_builtin_struct_get(candy_value_t *structure, candy_value_t *key)
+{
+    size_t index = 0;
+    while (!candy_tag_to_bool(candy_builtin_equals(structure->value.structure.keys[index], key)))
+    {
+        if (structure->value.structure.keys[index] == NULL)
+        {
+            candy_panic(make_candy_text("Attempted to access non-existent struct member"));
+        }
```
JonasWanke: This shouldn't be necessary because the wrapper in Candy checks that the key exists before calling this builtin

```suggestion
```

compiler/backend_inkwell/candy_rt/candy_builtin.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+candy_value_t *candy_builtin_struct_get_keys(candy_value_t *structure)
+{
+    return make_candy_list(structure->value.structure.keys);
+}
+
+const candy_value_t *candy_builtin_struct_has_key(candy_value_t *structure, candy_value_t *key)
+{
+
```
JonasWanke:
```suggestion
```

compiler/cli/src/main.rs
```diff
@@ -58,6 +62,7 @@ enum Exit {
     FuzzingFoundFailingCases,
     NotInCandyPackage,
     CodeContainsErrors,
+    LLVMError(String),
```
JonasWanke:
```suggestion
    LlvmError(String),
```

packages/Examples/int.candy
```diff
@@ -0,0 +1,3 @@
+x = 42
+
+main environment := x
```
JonasWanke: This can be removed in favor of `minimal.candy`

compiler/cli/src/inkwell.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// If enabled, the compiler will build the Candy runtime from scratch.
+    #[arg(long = "build-rt", default_value_t = false)]
+    build_rt: bool,
+    /// If enabled, the LLVM bitcode will be compiled with debug information.
+    #[arg(short = 'g', default_value_t = false)]
+    debug: bool,
+    /// The file or package to run. If none is provided, the package of your
+    /// current working directory will be run.
+    #[arg(value_hint = ValueHint::FilePath)]
+    path: Option<PathBuf>,
```
JonasWanke:
```suggestion
    /// If enabled, print the generated LLVM IR to stderr.
    #[arg(long = "print-llvm-ir", default_value_t = false)]
    print_llvm_ir: bool, 
    
    /// If enabled, print the output of the Candy main function.
    #[arg(long = "print-main-output", default_value_t = false)]
    print_main_output: bool,

    /// If enabled, build the Candy runtime from scratch.
    #[arg(long = "build-rt", default_value_t = false)]
    build_rt: bool,

    /// If enabled, compile the LLVM bitcode with debug information.
    #[arg(short = 'g', default_value_t = false)]
    debug: bool,

    /// The file or package to run. If none is provided, run the package of your
    /// current working directory.
    #[arg(value_hint = ValueHint::FilePath)]
    path: Option<PathBuf>,
```

compiler/cli/src/inkwell.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+pub(crate) fn compile(options: Options) -> ProgramResult {
+    let packages_path = packages_path();
+    let db = Database::new_with_file_system_module_provider(packages_path);
+    let path = options
+        .path
+        .as_ref()
+        .map_or_else(|| "Unknown".into(), |p| p.to_string_lossy().to_string());
```
JonasWanke: Is "Unknown" really correct here?

compiler/cli/src/inkwell.rs
```diff
@@ -0,0 +1,116 @@
...
+            if options.debug { "-g" } else { "" },
+            "-O3",
+            "-flto",
+            "-o",
+            &s_path.to_str().unwrap().replace(".candy.s", ""),
+        ])
+        .spawn()
+        .unwrap()
+        .wait()
+        .unwrap();
```
JonasWanke: I think it would be cleaner if this code lived in `backend_inkwell` itself, maybe with separate functions corresponding to the `print_*` flags

compiler/cli/src/inkwell.rs
```diff
@@ -0,0 +1,116 @@
...
+        .unwrap_or_else(|error| {
+            let payload = CompilerErrorPayload::Module(error);
+            let mir = Mir::build(|body| {
+                let reason = body.push_text(payload.to_string());
+                let responsible = body.push_hir_id(hir::Id::user());
+                body.push_panic(reason, responsible);
+            });
+            let errors = vec![CompilerError::for_whole_module(module.clone(), payload)]
+                .into_iter()
+                .collect();
```
JonasWanke:
```suggestion
            let errors =
                FxHashSet::from_iter([CompilerError::for_whole_module(module.clone(), payload)]);
```

compiler/cli/src/inkwell.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+        std::process::exit(1);
+    }
+
+    let context = backend_inkwell::inkwell::context::Context::create();
+    let module = context.create_module(&path);
+    let builder = context.create_builder();
+    let codegen = CodeGen::new(&context, module, builder, mir);
+    let mut bc_path = PathBuf::new();
+    bc_path.push(&format!("{path}.bc"));
```
JonasWanke:
```suggestion
    let bc_path = PathBuf::from(format!("{path}.bc"));
```

compiler/cli/src/inkwell.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        std::process::Command::new("make")
+            .args(["-C", "compiler/backend_inkwell/candy_rt/", "candy_rt.a"])
+            .spawn()
+            .unwrap()
+            .wait()
+            .unwrap();
+    }
+    let mut s_path = PathBuf::new();
+    s_path.push(&format!("{path}.s"));
```
JonasWanke:
```suggestion
    let s_path = PathBuf::from(format!("{path}.s"));
```

compiler/backend_inkwell/candy_rt/candy_rt.c
```diff
@@ -0,0 +1,204 @@
...
+        free_candy_value(length);
+        size_t index = 0;
+        for (size_t index = 0; index < list_length; index++)
+        {
+            print_candy_value(value->value.list[index]);
+            if (index != list_length - 1)
+            {
+                printf(", ");
+            }
+        }
```
JonasWanke:
```suggestion
        switch (list_length)
        {
        case 1:
            print_candy_value(value->value.list[0]);
        case 0:
            printf(",");
            break;
        default:
            for (size_t index = 0; index < list_length; index++)
            {
                print_candy_value(value->value.list[index]);
                if (index != list_length - 1)
                {
                    printf(", ");
                }
            }
            break;
        }
```

compiler/backend_inkwell/candy_rt/candy_rt.c
```diff
@@ -0,0 +1,204 @@
...
+const candy_value_t *to_candy_bool(int value)
+{
+    if (value)
+    {
+        return &__internal_true;
+    }
+    else
+    {
+        return &__internal_false;
+    }
```
JonasWanke:
```suggestion
    return value ? &__internal_true : &__internal_false;
```

compiler/backend_inkwell/candy_rt/candy_rt.c
```diff
@@ -0,0 +1,204 @@
...
+    candy_value_t *candy_value = malloc(sizeof(candy_value_t));
+    candy_value->type = CANDY_TYPE_STRUCT;
+    candy_value->value.structure.keys = keys;
+    candy_value->value.structure.values = values;
+    return candy_value;
+}
+
+candy_value_t *call_candy_function_with(candy_value_t *function, candy_value_t *arg)
+{
+    return function->value.function.function(arg);
```
JonasWanke: Doesn't this need the environment as well?

compiler/backend_inkwell/candy_rt/candy_rt.c
```diff
@@ -0,0 +1,204 @@
...
+    {
+        if (value->type == CANDY_TYPE_TAG || value->type == CANDY_TYPE_TEXT)
+        {
+            free(value->value.text);
+        }
+        // List and struct entries may not be freed as part of freeing
+        // the list/struct, because they will be freed on their own
+        // at the end of the main function.
+        free(value);
+    }
```
JonasWanke:
```suggestion
    if (value == candy_environment && value == NULL)
    {
        return;
    }

    if (value->type == CANDY_TYPE_TAG || value->type == CANDY_TYPE_TEXT)
    {
        free(value->value.text);
    }
    // List and struct entries may not be freed as part of freeing
    // the list/struct, because they will be freed on their own
    // at the end of the main function.
    free(value);
```

JonasWanke (changes requested):
I'm still working on `compiler/backend_inkwell/src/lib.rs`

compiler/backend_inkwell/candy_rt/candy_rt.c
ctiedt: This is currently used to call the candy main function which by definition cannot capture any environment. I might rename this function or add a comment to make this more clear.

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+    pub fn compile(
+        mut self,
+        path: &Path,
+        print_llvm_ir: bool,
+        print_main_output: bool,
+    ) -> Result<(), LLVMString> {
+        let i32_type = self.context.i32_type();
+        let i64_type = self.context.i64_type();
+        let i8_type = self.context.i8_type();
+        let void_type = self.context.void_type();
```
JonasWanke: Nit: I think this order is more intuitive

```suggestion
        let void_type = self.context.void_type();
        let i8_type = self.context.i8_type();
        let i32_type = self.context.i32_type();
        let i64_type = self.context.i64_type();
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+        let candy_type = self.context.opaque_struct_type("candy_type");
+        let candy_type_ptr = candy_type.ptr_type(AddressSpace::default());
```
JonasWanke: Maybe create one pointer type and add a comment about opaque pointers

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+        let struct_get_fn_type =
+            candy_type_ptr.fn_type(&[candy_type_ptr.into(), candy_type_ptr.into()], false);
+        let candy_builtin_struct_get =
+            self.module
+                .add_function("candy_builtin_struct_get", struct_get_fn_type, None);
+
+        let panic_fn_type = void_type.fn_type(
+            &[candy_type.ptr_type(AddressSpace::default()).into()],
+            false,
+        );
```
JonasWanke: For types that are reused (`make_tag_fn_type` as well), maybe a content-based name makes more sense, e.g., `pointer_to_void_function`? (Since it isn't used for panics only)

compiler/backend_inkwell/candy_rt/candy_rt.h
ctiedt: This doesn't work because of how C resolves types. It's fine with `struct candy_value`, but the typedef `candy_value_t` does not exist yet at this point.

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+    fn compile_mir(&mut self, mir: &Body, function_ctx: &FunctionInfo<'ctx>) {
+        let candy_type_ptr = self
+            .module
+            .get_struct_type("candy_type")
+            .unwrap()
+            .ptr_type(AddressSpace::default());
+
+        for (idx, (id, expr)) in mir.expressions.iter().enumerate() {
+            match expr {
+                candy_frontend::mir::Expression::Int(value) => {
```
JonasWanke: Can be imported

```suggestion
                Expression::Int(value) => {
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                    self.values.insert(*id, Rc::new(global));
+
+                    if idx == mir.expressions.len() - 1 {
+                        self.builder
+                            .build_return(Some(&global.as_basic_value_enum()));
+                    }
+                }
+                candy_frontend::mir::Expression::Text(text) => {
+                    let i32_type = self.context.i32_type();
+                    let i8_type = self.context.i8_type();
```
JonasWanke:
```suggestion
                    let i8_type = self.context.i8_type();
                    let i32_type = self.context.i32_type();
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+                    if idx == mir.expressions.len() - 1 {
+                        self.builder
+                            .build_return(Some(&global.as_basic_value_enum()));
+                    }
+                }
+                candy_frontend::mir::Expression::Tag { symbol, value } => {
+                    self.tags.insert(symbol.clone(), *value);
+                    let i32_type = self.context.i32_type();
+                    let i8_type = self.context.i8_type();
```
JonasWanke:
```suggestion
                    let i8_type = self.context.i8_type();
                    let i32_type = self.context.i32_type();
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+        self.compile_mir(&self.mir.body.clone(), &main_info);
+        self.builder.position_at_end(block);
+
+        let environment = self
+            .module
+            .add_global(candy_type_ptr, None, "candy_environment");
+
+        if let Some(main_return) = self.main_return {
+            const MAIN_FN_NAME: &str = "Main";
+            let len = i32_type.const_int(MAIN_FN_NAME.len() as u64, false);
```
JonasWanke: Shouldn't this be one longer byte to account for the null terminator?

```suggestion
            let len = i32_type.const_int(MAIN_FN_NAME.len() as u64 + 1, false);
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+            let main_text = self.builder.build_array_alloca(i8_type, len, "");
+            let main_text_array = self.make_str_literal(MAIN_FN_NAME);
+            self.builder.build_store(main_text, main_text_array);
+
+            let main_tag = self
+                .builder
+                .build_call(make_candy_tag, &[main_text.into()], "");
```
JonasWanke: Maybe extract a helper function for creating a tag since this is shared with `Expression::Tag(_)` below. And maybe a helper for just creating the C string which can be reused by `Expression::Text(_)`

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                    }
+                }
+                candy_frontend::mir::Expression::Tag { symbol, value } => {
+                    self.tags.insert(symbol.clone(), *value);
+                    let i32_type = self.context.i32_type();
+                    let i8_type = self.context.i8_type();
+
+                    let global = self.module.add_global(candy_type_ptr, None, symbol);
+                    let v = self.make_str_literal(symbol);
+                    let len = i32_type.const_int(symbol.len() as u64, false);
```
JonasWanke:
```suggestion
                    let len = i32_type.const_int(symbol.len() as u64 + 1, false);
```

compiler/backend_inkwell/candy_rt/candy_builtin.h
```diff
@@ -0,0 +1,17 @@
+#include "candy_rt.h"
+
+candy_value_t *candy_builtin_equals(candy_value_t *left, candy_value_t *right);
+candy_value_t *candy_builtin_ifelse(candy_value_t *condition, candy_value_t *then, candy_value_t *otherwise);
```
JonasWanke: For consistency:

```suggestion
candy_value_t *candy_builtin_if_else(candy_value_t *condition, candy_value_t *then, candy_value_t *otherwise);
```

compiler/backend_inkwell/candy_rt/candy_builtin.h
```diff
@@ -0,0 +1,17 @@
...
+candy_value_t *candy_builtin_int_add(candy_value_t *left, candy_value_t *right);
+candy_value_t *candy_builtin_int_subtract(candy_value_t *left, candy_value_t *right);
+candy_value_t *candy_builtin_int_bit_length(candy_value_t *value);
+candy_value_t *candy_builtin_int_bitwise_and(candy_value_t *left, candy_value_t *right);
+candy_value_t *candy_builtin_int_bitwise_or(candy_value_t *left, candy_value_t *right);
+candy_value_t *candy_builtin_int_bitwise_xor(candy_value_t *left, candy_value_t *right);
+candy_value_t *candy_builtin_int_compareto(candy_value_t *left, candy_value_t *right);
```
JonasWanke: For consistency:

```suggestion
candy_value_t *candy_builtin_int_compare_to(candy_value_t *left, candy_value_t *right);
```

compiler/backend_inkwell/candy_rt/candy_builtin.h
```diff
@@ -0,0 +1,17 @@
...
+candy_value_t *candy_builtin_list_length(candy_value_t *list);
+candy_value_t *candy_builtin_print(candy_value_t *value);
+candy_value_t *candy_builtin_struct_get(candy_value_t *structure, candy_value_t *key);
+candy_value_t *candy_builtin_struct_get_keys(candy_value_t *structure);
+candy_value_t *candy_builtin_struct_has_key(candy_value_t *structure, candy_value_t *key);
+candy_value_t *candy_builtin_typeof(candy_value_t *value);
```
JonasWanke: For consistency:

```suggestion
candy_value_t *candy_builtin_type_of(candy_value_t *value);
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                    self.values.insert(*id, Rc::new(global));
+
+                    if idx == mir.expressions.len() - 1 {
+                        self.builder
+                            .build_return(Some(&global.as_basic_value_enum()));
+                    }
+                }
+                candy_frontend::mir::Expression::Builtin(builtin) => {
+                    match builtin {
+                        candy_frontend::builtin_functions::BuiltinFunction::ChannelCreate => {
```
JonasWanke: Can be imported

```suggestion
                        BuiltinFunction::ChannelCreate => {
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                            .build_return(Some(&global.as_basic_value_enum()));
+                    }
+                }
+                candy_frontend::mir::Expression::Text(text) => {
+                    let i32_type = self.context.i32_type();
+                    let i8_type = self.context.i8_type();
+
+                    let global = self.module.add_global(candy_type_ptr, None, text);
+                    let v = self.make_str_literal(text);
+                    let len = i32_type.const_int(text.len() as u64 + 1, false);
```
JonasWanke: Shouldn't this use an `i64_type`?

compiler/backend_inkwell/src/lib.rs
JonasWanke: Shouldn't this use an `i64_type`?

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                candy_frontend::mir::Expression::List(list) => {
+                    let i32_type = self.context.i32_type();
+                    let candy_type_ptr = self
+                        .module
+                        .get_struct_type("candy_type")
+                        .unwrap()
+                        .ptr_type(AddressSpace::default());
+                    let list_array = self.builder.build_array_alloca(
+                        candy_type_ptr,
+                        i32_type.const_int(list.len() as u64 + 1, false),
```
JonasWanke: Shouldn't this use an `i64_type`?

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                        self.get_value_with_id(function_ctx, v)
+                            .unwrap()
+                            .into_pointer_value()
+                    });
+                    for (idx, value) in values.enumerate() {
+                        let value_position = unsafe {
+                            self.builder.build_gep(
+                                candy_type_ptr,
+                                list_array,
+                                &[i32_type.const_int(idx as u64, false)],
```
JonasWanke: Shouldn't this use an `i64_type`? Same for the `i32_type` occurrences below

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+                    let function_ctx_name =
+                        function_ctx.function_value.get_name().to_str().unwrap();
+                    if idx == mir.expressions.len() - 1 {
+                        if function_ctx_name != "main" {
+                            self.builder
+                                .build_return(Some(&struct_value.into_pointer_value()));
+                        } else {
+                            self.main_return.replace(struct_value.into_pointer_value());
+                        }
```
JonasWanke, line 498: There could be other functions called `main` returning a struct

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                                    function_value: Rc::new(function),
+                                    captured_ids: vec![],
+                                    env_type: None,
+                                },
+                            );
+                        }
+                    }
+                    self.unrepresented_ids.insert(*id);
+                }
+                candy_frontend::mir::Expression::List(list) => {
```
JonasWanke, line 364: I think this also has to check whether we're looking at the last expression and then call `self.builder.build_return(…)`

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+                    global.set_initializer(&candy_type_ptr.const_null());
+                    self.values.insert(*id, Rc::new(global));
+
+                    if idx == mir.expressions.len() - 1 {
+                        self.builder
+                            .build_return(Some(&global.as_basic_value_enum()));
+                    }
+                }
+                candy_frontend::mir::Expression::Builtin(builtin) => {
```
JonasWanke, line 325: I think this also has to check whether we're looking at the last expression and then call `self.builder.build_return(…)`

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+}
+
+pub struct CodeGen<'ctx> {
+    context: &'ctx Context,
+    module: Module<'ctx>,
+    builder: Builder<'ctx>,
+    mir: Arc<Mir>,
+    tags: HashMap<String, Option<Id>>,
+    values: HashMap<Id, Rc<GlobalValue<'ctx>>>,
+    locals: HashMap<Id, Rc<BasicValueEnum<'ctx>>>,
```
JonasWanke: I don't think the `Rc`s are necessary

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                        .module
+                        .get_struct_type("candy_type")
+                        .unwrap()
+                        .ptr_type(AddressSpace::default());
+
+                    let text = format!("{hir_id}");
+
+                    let global = self.module.add_global(candy_type_ptr, None, &text);
+                    let v = self.make_str_literal(&text);
+                    let len = i32_type.const_int(text.len() as u64, false);
```
JonasWanke:
```suggestion
                    let len = i32_type.const_int(text.len() as u64 + 1, false);
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                candy_frontend::mir::Expression::Function {
+                    original_hirs,
+                    parameters,
+                    body,
+                    responsible_parameter,
+                } => {
+                    let name = format!("{original_hirs:?}")
+                        .replace('{', "")
+                        .replace('}', "")
+                        .replace([':', '.'], "_");
```
JonasWanke: More explicit and with sorting:

```suggestion
                    let name = original_hirs
                        .iter()
                        .sorted()
                        .map(|it| it.to_string().replace([':', '.'], "_"))
                        .join(", ");
```

(Requires `itertools`)

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                    let make_candy_text = self.module.get_function("make_candy_text").unwrap();
+                    let call = self.builder.build_call(make_candy_text, &[cast.into()], "");
+
+                    self.builder.build_store(
+                        global.as_pointer_value(),
+                        call.try_as_basic_value().unwrap_left(),
+                    );
+
+                    global.set_initializer(&candy_type_ptr.const_null());
+                    self.values.insert(*responsible_parameter, Rc::new(global));
```
JonasWanke: This should instead use the value of the responsible parameter

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                            .build_struct_gep(env_struct_type, env_ptr, idx as u32, "")
+                            .unwrap();
+                        self.builder.build_store(member, value.unwrap());
+                    }
+
+                    let mut params: Vec<_> =
+                        parameters.iter().map(|_| candy_type_ptr.into()).collect();
+                    if !captured_ids.is_empty() {
+                        params.push(candy_type_ptr.into());
+                    }
```
JonasWanke:
```suggestion
                    let parameter_count = parameters.len() + usize::from(!captured_ids.is_empty());
                    let params = vec![candy_type_ptr.into(); parameter_count];
```

Plus another one for the responsible HIR ID

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+                    let fn_type = candy_type_ptr.fn_type(&params, false);
+
+                    let function = self.module.add_function(&name, fn_type, None);
+
+                    let fun_info = FunctionInfo {
```
JonasWanke: For consistency

```suggestion
                    let function_info = FunctionInfo {
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                        self.module
+                            .add_global(candy_type_ptr, None, &format!("fun_{name}"));
+                    global.set_initializer(&function_ptr.get_type().const_null());
+
+                    self.builder.build_store(
+                        global.as_pointer_value(),
+                        call.try_as_basic_value().unwrap_left(),
+                    );
+
+                    self.values.insert(*id, Rc::new(global));
```
JonasWanke: Can you maybe extract this into a helper function, since it looks like it's done in multiple places? Also, please rename `self.values` to `globals` or `constants`

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                            self.builder
+                                .build_call(get_candy_fn_ptr, &[function_value.into()], "");
+
+                        let fn_env_ptr =
+                            self.builder
+                                .build_call(get_candy_fn_env, &[function_value.into()], "");
+
+                        args.push(fn_env_ptr.try_as_basic_value().unwrap_left().into());
+
+                        let candy_fn_type = candy_type_ptr.fn_type(&[], true);
```
JonasWanke: Shouldn't this be the actual function type with the correct number of parameters instead of using var args?

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                        }
+                    }
+                }
+                candy_frontend::mir::Expression::UseModule { .. } => unreachable!(),
+                candy_frontend::mir::Expression::Panic { reason, .. } => {
+                    let panic_fn = self.module.get_function("candy_panic").unwrap();
+
+                    let reason = self.get_value_with_id(function_ctx, reason).unwrap();
+
+                    self.builder.build_call(panic_fn, &[reason.into()], "");
```
JonasWanke, line 708: It would be good to print the responsible HIR ID as well

compiler/backend_inkwell/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "candy_backend_inkwell"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+candy_frontend = { version = "0.1.0", path = "../frontend" }
+inkwell = { version = "0.2.0", features = ["llvm15-0"] }
+llvm-sys = { version = "150", features = ["prefer-dynamic"] }
```
JonasWanke:
```suggestion
llvm-sys = { version = "150", features = ["prefer-dynamic"] }

```

compiler/backend_inkwell/candy_rt/candy_rt.c
```diff
@@ -0,0 +1,209 @@
...
+{
+    printf("The program panicked for the following reason: \n");
+    print_candy_value(reason);
+    printf("\n");
+    exit(-1);
+}
+
+void free_candy_value(candy_value_t *value)
+{
+    if (value == candy_environment && value == NULL)
```
JonasWanke:
```suggestion
    if (value == candy_environment || value == NULL)
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+                    .unwrap();
+
+                v.replace(self.builder.build_load(candy_type_ptr, env_value, ""));
+            }
+        }
+        if v.is_none() {
+            if let Some(value) = self.locals.get(id) {
+                v.replace(*value.clone());
+            }
+        }
```
JonasWanke: With `#![feature(let_chains)]`:

```suggestion
        if v.is_none() && let Some(i) = function_ctx.captured_ids.iter().position(|i| i == id) {
            let env_ptr = function_ctx.function_value.get_last_param().unwrap();

            let env_value = self
                .builder
                .build_struct_gep(
                    function_ctx.env_type.unwrap(),
                    env_ptr.into_pointer_value(),
                    i as u32,
                    "",
                )
                .unwrap();

            v = Some(self.builder.build_load(candy_type_ptr, env_value, ""));
        }
        if v.is_none() && let Some(value) = self.locals.get(id) {
            v = Some(*value.clone());
        }
```

JonasWanke (changes requested):

compiler/backend_inkwell/src/lib.rs
ctiedt: Will be fixing this in a follow-up issue

compiler/backend_inkwell/src/lib.rs
ctiedt: Since this is the path for an indirect function call, we don't know how many parameters the function has (since functions with different signatures are saved as the same type). It might be possible to save that information in the function object, but extracting it would incur further FFI costs and the compiler should have verified at an earlier stage that the number of parameters is correct.

compiler/backend_inkwell/src/lib.rs
ctiedt: I decided to remove responsible IDs for now and will revisit the topic later.

compiler/backend_inkwell/src/lib.rs
ctiedt: As long as inkwell does not properly support opaque pointers, I think leaving the code like this is more readable. Will change this as soon as inkwell represents opaque pointers correctly.

compiler/backend_inkwell/src/lib.rs
ctiedt: Will be addressed when working on #629

compiler/backend_inkwell/src/lib.rs
ctiedt: Will be addressed when working on #629

compiler/backend_inkwell/candy_runtime/candy_builtin.c
```diff
@@ -41,7 +42,14 @@ candy_value_t *candy_builtin_int_subtract(candy_value_t *left, candy_value_t *ri
 
 candy_value_t *candy_builtin_int_bit_length(candy_value_t *value)
 {
-    return make_candy_int(64);
+    long long int_value = value->value.integer;
```
JonasWanke:
```suggestion
    int64_t int_value = value->value.integer;
```

compiler/backend_inkwell/candy_runtime/candy_builtin.c
```diff
@@ -41,7 +42,14 @@ candy_value_t *candy_builtin_int_subtract(candy_value_t *left, candy_value_t *ri
...
+    int shifts = 0;
+    while (int_value)
+    {
+        int_value = int_value >> shifts;
+        shifts++;
+    }
```
JonasWanke, line 56: I don't think this works correctly for negative numbers

compiler/cli/src/inkwell.rs
```diff
@@ -0,0 +1,97 @@
...
+    #[arg(long = "print-llvm-ir", default_value_t = false)]
+    print_llvm_ir: bool,
+
+    /// If enabled, print the output of the Candy main function.
+    #[arg(long = "print-main-output", default_value_t = false)]
+    print_main_output: bool,
+
+    /// If enabled, build the Candy runtime from scratch.
+    #[arg(long = "build-rt", default_value_t = false)]
+    build_rt: bool,
```
JonasWanke: For consistency

```suggestion
    #[arg(long = "build-runtime", default_value_t = false)]
    build_runtime: bool,
```

compiler/cli/src/inkwell.rs
```diff
@@ -67,51 +82,16 @@ pub(crate) fn compile(options: Options) -> ProgramResult {
...
 
-    let context = backend_inkwell::inkwell::context::Context::create();
-    let module = context.create_module(&path);
-    let builder = context.create_builder();
-    let codegen = CodeGen::new(&context, module, builder, mir);
-    let bc_path = PathBuf::from(format!("{path}.bc"));
+    let context = candy_backend_inkwell::inkwell::context::Context::create();
+    let mut codegen = CodeGen::new(&context, &path, mir);
+    let mut bc_path = PathBuf::new();
+    bc_path.push(&format!("{path}.bc"));
```
JonasWanke: The same is built in `compile_asm_and_link(…)`

compiler/backend_inkwell/src/lib.rs
```diff
@@ -281,290 +306,69 @@ impl<'ctx> CodeGen<'ctx> {
...
-                        global.as_pointer_value(),
-                        call.try_as_basic_value().unwrap_left(),
-                    );
+                    let global =
+                        self.create_global(symbol, id, call.try_as_basic_value().unwrap_left());
 
-                    global.set_initializer(&candy_type_ptr.const_null());
-                    self.values.insert(*id, Rc::new(global));
+                    global.set_initializer(&candy_value_ptr.const_null());
+                    self.globals.insert(*id, global);
```
JonasWanke:
```suggestion
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -0,0 +1,800 @@
...
+        let candy_value_ptr = self
+            .module
+            .get_struct_type("candy_value")
+            .unwrap()
+            .ptr_type(AddressSpace::default());
+        let global = self.module.add_global(candy_value_ptr, None, name);
+        self.builder.build_store(global.as_pointer_value(), value);
+
+        global.set_initializer(&candy_value_ptr.const_null());
+        self.globals.insert(*id, global);
```
JonasWanke:
```suggestion
        assert!(self.globals.insert(*id, global).is_none());
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -864,36 +598,37 @@ impl<'ctx> CodeGen<'ctx> {
...
                 } => {
+                    self.unrepresented_ids.insert(*responsible);
                     let mut args: Vec<_> = arguments
                         .iter()
                         .map(|arg| self.get_value_with_id(function_ctx, arg).unwrap().into())
                         .collect();
+
                     if let Some(FunctionInfo {
                         function_value,
+
```
JonasWanke:
```suggestion
```

compiler/backend_inkwell/src/lib.rs
```diff
@@ -989,51 +724,75 @@ impl<'ctx> CodeGen<'ctx> {
         }
     }
 
-    fn make_str_literal(&self, s: &str) -> ArrayValue<'_> {
+    fn create_global<V: BasicValue<'ctx>>(
+        &mut self,
+        name: &str,
+        id: &Id,
+        value: V,
```
JonasWanke:
```suggestion
    fn create_global(
        &mut self,
        name: &str,
        id: &Id,
        value: impl BasicValue<'ctx>,
```

JonasWanke (changes requested):
Oh, and please add the new crate to `/compiler.code-workspace`, maybe below the VM

JonasWanke (approved):

PR #610, opened 2023-08-05
New LIR

JonasWanke:
Why just have one LIR when you can have two‽

### Checklist


compiler/frontend/src/mir_optimize/reference_following.rs
```diff
@@ -28,15 +28,8 @@ pub fn follow_references(context: &mut Context, expression: &mut CurrentExpressi
...
-            unreachable!()
-        };
-
-        if let Expression::Reference(referenced) = last_expression && referenced == second_last_id {
-            body.expressions.pop();
-        } else {
-            break;
-        }
+    while let [.., (second_last_id, _), (_, Expression::Reference(referenced))] = &body.expressions[..]
+        && referenced == second_last_id {
```
MarcelGarus, line 32: Would this also work?

```suggestion
    while let [.., (id, _), (_, Expression::Reference(id))] = &body.expressions[..] {
```

compiler/frontend/src/lir/body.rs
```diff
@@ -0,0 +1,200 @@
...
+/// - parameters
+/// - responsible parameter
+/// - locals
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Body {
+    original_hirs: FxHashSet<hir::Id>,
+    captured_count: usize,
+    parameter_count: usize,
+    expressions: Vec<Expression>,
+}
```
MarcelGarus, line 116: I get that it's _possible_ to not save the IDs because they can all be implicitly calculated. I imagine this will make writing optimization passes much harder though. But I guess that's a problem for future-us.

compiler/frontend/src/mir_to_lir.rs
```diff
@@ -0,0 +1,391 @@
...
+    Ok((Arc::new(lir), errors))
+}
+
+#[derive(Clone, Debug, Default)]
+struct LoweringContext {
+    constants: lir::Constants,
+    constant_mapping: FxHashMap<mir::Id, lir::ConstantId>,
+    bodies: lir::Bodies,
+    current_body: CurrentBody,
+    last_constant: Option<mir::Id>,
```
MarcelGarus: I think we can move some of this around so that `mem::replace` is no longer necessary. What if the `LoweringContext` contained only things where data is never removed (`constants`, `bodies`, `constant_mapping`) and the `last_constant` is moved into the `CurrentBody` and set during `finish`?

MarcelGarus (changes requested):

compiler/frontend/src/mir_optimize/reference_following.rs
JonasWanke: I tried this, but Rust complained about two captures with the same name

compiler/frontend/src/lir/body.rs
JonasWanke: Yeah, we can change that if it becomes necessary

MarcelGarus (approved):

PR #613, opened 2023-08-06
Newline support in text

### LINKED ISSUE #561: Fix text parsing and formatting
Parsing and formatting of texts should be similar to parsing of structs in that the `"`s should only be on the same line as the text content if it fits on a single line. Otherwise, they should be formatted like this:

```candy
foo = "
  Some text
  That is long
"
```

Also, we currently trim all newlines. The newlines should still exist in the text so that what's in the editor pretty much exactly matches what you get in Candy (except for the indentation).

skyz1:
Closes: #561 

Newlines in text do not get trimmed anymore. This PR also adds some basic text support to the formatter.

### Checklist


MarcelGarus (changes requested):
Very nice!

Can we remove the `TextLine`? Theoretically, a text can still be represented as a bunch of `TextPart`s and interpolation values with with trailing whitespace.

skyz1:
> Can we remove the `TextLine`? Theoretically, a text can still be represented as a bunch of `TextPart`s and interpolation values with with trailing whitespace.

You are right, `TextLine` is not absolutely necessary. However, without `TextLine` the code in other places would become more complex in other places, since line breaks inside text cannot be treated as whitespace without semantic meaning like everywhere else. I tried it in the beginning, but it turned out to be cumbersome.

I agree that the approach of introducing `TextLine` is not ideal. If you have an idea that doesn't require treating whitespace inside text differently, I'd be happy to implement it.

MarcelGarus:
@skyz1 Okay, fair enough. Then I'm fine with how it's implemented.

MarcelGarus (approved):

skyz1:
> @skyz1 Okay, fair enough. Then I'm fine with how it's implemented.

@MarcelGarus Before we merge this PR I want to propose two alternative solutions:
1. Introduce `TextNewline` that is similar to `Newline`, but is only used in `Text` to convey that line breaks in text are not whitespace without semantic meaning.
2. Always insert newlines between two consecutive `TextPart`s when lowering the CST. This would require inserting empty TextPart's for empty lines or before/after a `TextInterpolation` at the start/end of a line.

Do you think one of those proposals is better then the current solution using `TextLine`?

MarcelGarus:
@skyz1 Nice proposal, I like that even better.

Building on that, I could also imagine an approach where a `Text` would contain both `TextPart`s and `TextNewline`s so that empty `TextPart`s are no longer necessary.
Something like this:

```candy
"
  foo
  
  bar
"
```

```
CST:
Text {
  opening_quotes: TrailingWhitespace {
    child: DoubleQuote,
    whitespace: [Newline, Whitespace("  ")],
  },
  parts: [
    TextPart("foo"),
    TrailingWhitespace {
      child: TextNewline,
      whitespace: [Whitespace("  ")],
    },
    TrailingWhitespace {
      child: TextNewline,
      whitespace: [Whitespace("  ")],
    },
    TrailingWhitespace {
      child: TextPart("bar"),
      whitespace: [Newline],
    },
  ],
  closing_quote: DoubleQuote,
}
```

```
AST:
Text {
  parts: [
    TextPart("foo"),
    TextNewline,
    TextNewline,
    TextPart("bar"),
  ],
}
```

But I'm also okay with empty `TextPart`s that have `TextNewline`s as trailing whitespace.

MarcelGarus:
Very nice!

MarcelGarus (approved):

PR #650, opened 2023-08-12
Remove symbol table

### LINKED ISSUE #609: Remove Symbol Table
Follow-up for #590 

Alternative Idea: We revert symbols to having a pointer to a text instead of a symbol ID. We can still have an inline representation of symbols (#598), and since texts are deduplicated and (at least for symbols) statically known, the performance will be the same. However, we can reduce code complexity since we no longer have to pass around a symbol table. To allow efficient creation of symbols/tags in builtin functions, the heap could store pointers to these around 20 predefined texts.

JonasWanke:
Closes: #609 

### Checklist


compiler/fuzzer/src/input.rs
```diff
@@ -14,18 +14,14 @@ pub struct Input {
...
             f,
             "{}",
             self.arguments
                 .iter()
-                .map(|argument| argument.to_debug_text(
-                    Precedence::High,
-                    MaxLength::Limited(40),
-                    symbol_table
-                ))
+                .map(|argument| argument.to_debug_text(Precedence::High, MaxLength::Limited(40),))
```
MarcelGarus:
```suggestion
                .map(|argument| argument.to_debug_text(Precedence::High, MaxLength::Limited(40)))
```

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #668, opened 2023-08-17
Document built-ins

MarcelGarus:
Documents the built-ins and makes them more consistent.

### Checklist


compiler/frontend/src/builtin_functions.rs
```diff
@@ -7,53 +7,413 @@ use lazy_static::lazy_static;
...
+    /// `False`.
+    ///
+    /// TODO: Document how equality works. Also, change how equality for
+    /// functions works.
+    ///
+    /// ```candy
+    /// ✨.equals 5 5 => True
+    /// ✨.equals 5 2 => False
+    /// ```
+    Equals,
```
JonasWanke: I think it makes more sense to move all of these comments over to the Builtins package (`packages/Builtins/_.candy`, where some builtins already have documentation)

compiler/vm/src/builtin_functions.rs
```diff
@@ -266,9 +266,11 @@ impl Heap {
...
     fn int_parse(&mut self, args: &[InlineObject]) -> BuiltinResult {
         unpack_and_later_drop!(self, args, |text: Text| {
-            let result = match BigInt::from_str(text.get()) {
-                Ok(int) => Ok(Int::create_from_bigint(self, true, int).into()),
-                Err(err) => Err(Text::create(self, true, &ToString::to_string(&err)).into()),
+            let result = if let Ok(int) = BigInt::from_str(text.get()) {
+                Ok(Int::create_from_bigint(self, true, int).into())
+            } else {
+                let not_an_integer = Text::create(self, true, "NotAnInteger");
+                Err(Tag::create_with_value_option(self, true, not_an_integer, None).into())
```
JonasWanke:
```suggestion
                Err(Tag::create(self, not_an_integer).into())
```

However, I suggest adding the actual text we couldn't parse in there:

```suggestion
                Err(Tag::create_with_value(self, true, not_an_integer, text).into())
```

(Plus proper duping)

compiler/vm/src/builtin_functions.rs
```diff
@@ -266,9 +266,11 @@ impl Heap {
     }
     fn int_parse(&mut self, args: &[InlineObject]) -> BuiltinResult {
         unpack_and_later_drop!(self, args, |text: Text| {
-            let result = match BigInt::from_str(text.get()) {
-                Ok(int) => Ok(Int::create_from_bigint(self, true, int).into()),
-                Err(err) => Err(Text::create(self, true, &ToString::to_string(&err)).into()),
+            let result = if let Ok(int) = BigInt::from_str(text.get()) {
+                Ok(Int::create_from_bigint(self, true, int).into())
+            } else {
+                let not_an_integer = Text::create(self, true, "NotAnInteger");
```
JonasWanke: This could be moved into `DefaultSymbols`

compiler/vm/src/heap/object.rs
```diff
@@ -448,7 +448,10 @@ impl Text {
     pub fn create_from_utf8(heap: &mut Heap, is_reference_counted: bool, bytes: &[u8]) -> Tag {
         let result = str::from_utf8(bytes)
             .map(|it| Self::create(heap, is_reference_counted, it).into())
-            .map_err(|_| Self::create(heap, is_reference_counted, "Invalid UTF-8.").into());
+            .map_err(|_| {
+                let not_utf8 = Self::create(heap, true, "NotUtf8");
+                Tag::create_with_value_option(heap, true, not_utf8, None).into()
```
JonasWanke: Comments from `int_parse` apply here, too

JonasWanke (changes requested):

JonasWanke (approved):

packages/Builtins/_.candy
```diff
@@ -37,22 +37,51 @@ isLessThanOrEqualTo a b =
     Greater -> False
 
 equals a b :=
-  # Returns `True` or `False` depending on whether `a` and `b` are equal.
+  # Returns `True` if both values are equal, otherwise `False`.
+  #
+  # TODO: Document how equality works. Also, change how equality for
```
ctiedt: Please actually add how equality is defined in Candy as the todo suggests.

packages/Builtins/_.candy
```diff
@@ -37,22 +37,51 @@ isLessThanOrEqualTo a b =
...
+  #
+  # ```
+  # getArgumentCount { a -> a } => 1
+  # getArgumentCount getArgumentCount => 1
+  # ```
   needs (function | typeIs Function)
   ✨.getArgumentCount function
 
 ifElse condition then else :=
+  # If the condition is `True`, runs `then`. If it's `False`, runs `else`.
```
ctiedt, line 123:
```suggestion
  # If the condition evaluates to `True`, runs `then`. If it's `False`, runs `else`.
```

packages/Builtins/_.candy
```diff
@@ -77,164 +106,382 @@ fitsInRustU128 value =
...
+  # ```
   needs (a | typeIs Int)
   needs (b | typeIs Int)
   ✨.intCompareTo a b
+
 intDivideTruncating dividend divisor :=
   # Returns `dividend` ÷ `divisor`.
   #
-  # This operation rounds towards zero, truncating any fractional part of the exact result.
+  # This operation rounds towards zero, truncating any fractional part of the
```
ctiedt: Maybe mention that this operation is for division. For all other operations, this is mentioned first and might make it sound like this builtin is for rounding.

packages/Builtins/_.candy
```diff
@@ -77,164 +106,382 @@ fitsInRustU128 value =
...
+  # intDivideTruncating 6 3 => 2
+  # intDivideTruncating 5 3 => 1
+  # ```
   needs (dividend | typeIs Int)
   needs (divisor | typeIs Int)
   needs (divisor | equals 0 | not) "You can't divide by zero."
   ✨.intDivideTruncating dividend divisor
+
 intModulo dividend divisor :=
+  # Returns `modulo` % `divisor`.
```
ctiedt:
```suggestion
  # Returns `dividend` % `divisor`.
```

Also, it might be worth mentioning how you deal with negative numbers here.

packages/Builtins/_.candy
```diff
@@ -77,164 +106,382 @@ fitsInRustU128 value =
...
   needs (factorA | typeIs Int)
   needs (factorB | typeIs Int)
   ✨.intMultiply factorA factorB
+
 intParse text :=
-  # Parses `text` as an integer.
+  # Parses `text` into an integer.
+  # 
+  # `text` must be a string of decimal digits, optionally preceded by a minus
+  # sign. If the text is the textual representation of an integer, returns `Ok`
```
ctiedt, line 266:
```suggestion
  # sign. If the text is the textual representation of an integer, returns
```

packages/Builtins/_.candy
```diff
@@ -77,164 +106,382 @@ fitsInRustU128 value =
...
+  # listFilled 3 Foo => (Foo, Foo, Foo)
+  # listFilled 10 1 => (1, 1, 1, 1, 1, 1, 1, 1, 1, 1)
+  # listFilled 0 Bar => (,)
+  # ```
   needs (length | typeIs Int)
   needs (length | isNonNegative)
   ✨.listFilled length item
+
 listGet list index :=
+  # Returns the item that's at the `index` in the `list`.
```
ctiedt:
```suggestion
  # Returns the item at position `index` in the `list`.
```

packages/Builtins/_.candy
```diff
@@ -77,164 +106,382 @@ fitsInRustU128 value =
...
+  #
+  # ```
+  # textConcatenate "Hello" "world" => "Helloworld"
+  # ```
   needs (a | typeIs Text)
   needs (b | typeIs Text)
   ✨.textConcatenate a b
+
 textContains text pattern :=
+  # Returns whether the `text` contains the `pattern`.
```
ctiedt: Is the pattern a literal substring or a pattern like e.g. regex?

ctiedt (changes requested):

packages/Builtins/_.candy
MarcelGarus: From the view of the `ifElse` function, the `condition` parameter should just _be_ `True` – the expression is evaluated before the call even happens. So I'd actually leave it as-is.

ctiedt (approved):

ctiedt (approved):

JonasWanke (approved):

PR #645, opened 2023-08-12
Treat responsibilities like other parameters

MarcelGarus:
Closes: #630

### Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

PR #669, opened 2023-08-17
Guard the calls with checks

MarcelGarus:
Depends on #645

This PR makes it so that `Call` instruction can rely on more stuff:

- the callee is not a tag (there's now a `tagWithValue` builtin that associates a tag with a value)
- `✨.typeOf callee` is `Function`
- the number of arguments matches the expected count

### Checklist


JonasWanke:
Can this be closed since it's superseded by #909?

PR #772, opened 2023-10-25
Chrono package

### LINKED ISSUE #732: Chrono Package

skyz1:
Closes: #732

This PR adds a basic `Chrono` package that supports:
1. Instants
2. Accessing the System Time
3. Durations

### Checklist


packages/Chrono/clock.candy
```diff
@@ -0,0 +1,21 @@
...
+    systemClock | function.run | duration.fromNanoseconds | instant.fromUnixDuration
+  }
+
+mockFromInstant inst := 
+  needs (instant.is inst)
+  Clock { instant }
+
+now clock :=
+  needs (is clock)
+  function.run (tag.getValue clock)
```
JonasWanke:
```suggestion
  clock | tag.getValue | function.run
```

packages/Chrono/duration.candy
```diff
@@ -0,0 +1,198 @@
+[fixedDecimal, tag, check, equals, bool, ifElse, int, toDebugText, text] = use "Core"
+constants = use "..constants"
+
+# Storage Format: Duration (fixedDecimal of nanoseconds)
```
JonasWanke: Single octothorpes are doc comments that apply to the item above

```suggestion
## Storage Format: Duration (fixedDecimal of nanoseconds)
```

JonasWanke: Instead of starting fractions after the nanoseconds, I'd find a fixed decimal of seconds more intuitive here

packages/Chrono/duration.candy
```diff
@@ -0,0 +1,198 @@
...
+  needs (isNumber minutes)
+  Duration (numberAsFixedDecimal minutes | fixedDecimal.multiply (fixedDecimal.fromInt constants.nanosecondsPerMinute))
+fromHours hours :=
+  needs (isNumber hours)
+  Duration (numberAsFixedDecimal hours | fixedDecimal.multiply (fixedDecimal.fromInt constants.nanosecondsPerHour))
+fromNormalDays days :=
+  needs (isNumber days)
+  Duration (numberAsFixedDecimal days | fixedDecimal.multiply (fixedDecimal.fromInt constants.nanosecondsPerNormalDay))
+
+# TODO: Do this without truncating division if possible
```
JonasWanke:
```suggestion
## TODO: Do this without truncating division if possible
```

packages/Chrono/instant.candy
```diff
@@ -0,0 +1,78 @@
...
+  Instant nanoseconds -> fixedDecimal.is nanoseconds
+  _ -> False
+
+startOfUnixEpoch := Instant fixedDecimal.zero
+
+fromUnixDuration dur :=
+  needs (duration.is dur)
+  Instant (tag.getValue dur)
+
+addDuration instant dur :=
```
JonasWanke:
```suggestion
add instant dur :=
```

packages/Chrono/instant.candy
```diff
@@ -0,0 +1,78 @@
...
+  Instant (fixedDecimal.add (tag.getValue instant) (tag.getValue dur))
+subtract instant dur :=
+  needs (is instant)
+  needs (duration.is dur)
+  Instant (fixedDecimal.subtract (tag.getValue instant) (tag.getValue dur))
+
+durationBetween instantA instantB :=
+  needs (is instantA)
+  needs (is instantB)
+  duration.fromNanoseconds (fixedDecimal.subtract (tag.getValue instantA) (tag.getValue instantB))
```
JonasWanke: We could merge these both to a single `subtract` function that acts depending on the value of the second parameter

packages/Chrono/instant.candy
```diff
@@ -0,0 +1,78 @@
+[fixedDecimal, tag, check, equals, bool, ifElse] = use "Core"
+constants = use "..constants"
+duration = use "..duration"
+
+# Storage Format: Instant (fixedDecimal of nanoseconds since the Unix Epoch)
```
JonasWanke: What do you think about storing a `duration since the unix expoch` inside instead of a raw fixed decimal?

(Also, two octothorpes here)

packages/Core/text.candy
```diff
@@ -120,3 +120,29 @@ trimEnd := builtins.textTrimEnd
...
+
+padStart text len char :=
+  needs (is text)
+  needs (int.is len)
+  needs (is char)
+  needs (char | length | equals 1)
+  recursive text { recurse text ->
+    ifElse (text | length | int.isGreaterThanOrEqualTo len) {text} {char | concatenate text | recurse}
+  }
+
```
JonasWanke: These two functions belong together, so we can group by removing this empty line

```suggestion
```

packages/Core/text.candy
```diff
@@ -120,3 +120,29 @@ trimEnd := builtins.textTrimEnd
...
+padEnd text len char :=
+  needs (is text)
+  needs (int.is len)
+  needs (is char)
+  needs (char | length | equals 1)
+  recursive text { recurse text ->
+    ifElse (text | length | int.isGreaterThanOrEqualTo len) {text} {text | concatenate char | recurse}
+  }
```
JonasWanke: We could add a function `text.repeat text times` as a reusable helper

packages/Examples/clock.candy
```diff
@@ -0,0 +1,10 @@
+[clock, duration, instant] = use "Chrono"
+[tag, fixedDecimal] = use "Core"
+
+main := { environment ->
+  print message = environment.stdout message
+
+  c = clock.fromSystemClock environment.systemClock
+  now = clock.now c
+  print "It has been {now | instant.toUnixTimeDuration | duration.toText} since the start the of unix epoch"
```
JonasWanke:
```suggestion
  print "It has been {now | instant.toUnixTimeDuration | duration.toText} since the start the of Unix epoch"
```

JonasWanke (changes requested):

packages/Chrono/instant.candy
skyz1: I think it is better to keep those functions separated since the first returns an `Instant` and the second a `Duration`. I think it would be confusing if a function returned values of different types depending on the argument types.

packages/Chrono/instant.candy
skyz1: I agree with you, but this approach would be slower. Either we implement your suggestion and hope that the compiler can optimize such type tags in the future or we use the faster variant for the time being and leave a note.

JonasWanke: Speed is not that important right now and we still have a few compiler optimizations planned, so I'd go with the cleaner, albeit slower approach

packages/Chrono/instant.candy
JonasWanke: Having both in the same function basically corresponds to overloading in languages like Kotlin or Java, and as a developer looking for either of those functions, `subtract` should be intuitive. I'd like to try merging these two and see if (also together with the tooling showing function results live in the editor) this suffices, and we can use that pattern in general. Otherwise, we'll revert it to two separate functions in the future.

skyz1:
@JonasWanke Sorry for the delay, I've been very busy lately. I have addressed your remaining comments.

packages/Chrono/instant.candy
```diff
@@ -75,4 +72,4 @@ coerceIn value minimum maximum :=
 
 toUnixTimeDuration instant :=
   needs (is instant)
-  duration.fromNanoseconds (tag.getValue instant)
+  duration.fromSeconds (tag.getValue instant)
```
JonasWanke: The value is already a duration

```suggestion
  tag.getValue instant
```

JonasWanke (changes requested):
I found one bug, but the rest looks good

JonasWanke (approved):

PR #861, opened 2024-01-06
Implement match condition

### LINKED ISSUE #659: Implement match condition
Match cases should have an optional condition that also has to be fulfilled for the case to be executed.

https://github.com/candy-lang/candy/blob/e98f2c346128138aed6b29f363237d8819f2cd59/language.md#pattern-matching

fpottbaecker:
Closes: #659 

Added optional condtions to match cases in RCST, AST, and HIR, which are resolved to if/else in MIR.

### Checklist


compiler/frontend/src/hir_to_mir.rs
```diff
@@ -549,55 +521,111 @@ impl<'a> LoweringContext<'a> {
...
-                    let list_get_function = body.push_builtin(BuiltinFunction::ListGet);
-                    let one = body.push_int(1);
-                    let reason = body.push_call(
-                        list_get_function,
-                        vec![pattern_result, one],
-                        responsible_for_match,
-                    );
-                    no_match_reasons.push(reason);
 
+                let else_function = body.push_function(case_id.child("didNotMatch"), |body, _| {
```
fpottbaecker, line 653: Note: The else function is now pushed before the then function, which makes reading the MIR a little more convoluted.

compiler/formatter/src/format.rs
```diff
@@ -744,6 +744,7 @@ pub fn format_cst<'a>(
         }
         CstKind::MatchCase {
             pattern,
+            condition: _, // TODO: format match case conditions
```
MarcelGarus, line 745: I opened a follow-up issue: #868

compiler/frontend/src/mir/body.rs
```diff
@@ -411,7 +411,53 @@ impl BodyBuilder {
             responsible,
         )
     }
+    pub fn push_if_not<E>(
+        &mut self,
+        hir_id: &hir::Id,
+        condition: Id,
+        else_builder: E,
```
MarcelGarus: You can inline the constraint so there are no type parameters anymore.
```suggestion
        else_builder: impl FnOnce(&mut Self),
```

MarcelGarus (changes requested):

compiler/frontend/src/rcst.rs
```diff
@@ -17,7 +17,7 @@ impl From<CstKind<()>> for Cst<()> {
 
 impl ToRichIr for Rcst {
     fn build_rich_ir(&self, builder: &mut RichIrBuilder) {
-        builder.push(format!("{self:?}"), None, EnumSet::empty());
+        builder.push(format!("{self:#?}"), None, EnumSet::empty());
```
JonasWanke: The CST's rich IR is currently messed up and I suspect it comes from this change

[candy-lang/golden-irs@`a68dbb9b012ede17ab5d72dc6cc793aef4cf42e0_..3e506fd0a8da852079633eb18f932fe0baaeca27_`#diff-ff8b15b562](https://github.com/candy-lang/golden-irs/compare/a68dbb9b012ede17ab5d72dc6cc793aef4cf42e0_..3e506fd0a8da852079633eb18f932fe0baaeca27_#diff-ff8b15b5625a7cf5381cc45750d5c17c95d699e3eaf7f0d1de2cdd29cf7dfa6e)

compiler/frontend/src/string_to_rcst/expression.rs
```diff
@@ -405,6 +405,31 @@ fn match_case(input: &str, indentation: usize) -> Option<(&str, Rcst)> {
     let (input, whitespace) = whitespaces_and_newlines(input, indentation, true);
     let pattern = pattern.wrap_in_whitespace(whitespace);
 
+    let (input, condition) = if let Some((input, condition_comma)) = comma(input) {
+        let (input, whitespace) = whitespaces_and_newlines(input, indentation, true);
+        let condition_comma = condition_comma.wrap_in_whitespace(whitespace);
+        if let Some((input, condition_expresion)) = expression(
```
JonasWanke: Small typo: `condition_expresion` is missing one “s”

compiler/frontend/src/string_to_rcst/expression.rs
```diff
@@ -429,6 +454,7 @@ fn match_case(input: &str, indentation: usize) -> Option<(&str, Rcst)> {
 
     let case = CstKind::MatchCase {
         pattern: Box::new(pattern),
+        condition: condition.map(Box::new),
```
JonasWanke, line 459: Can you please add a test case that contains a match case with a condition?

compiler/frontend/src/hir_to_mir.rs
```diff
@@ -385,6 +357,8 @@ impl<'a> LoweringContext<'a> {
                     self.ongoing_destructuring.clone().unwrap();
 
                 if is_trivial {
+                    // something %
+                    //   foo -> ...
```
JonasWanke:
```suggestion
                    // foo = …
                    //
                    // or:
                    //
                    // something %
                    //   foo -> ...
```

compiler/frontend/src/mir/body.rs
```diff
@@ -411,7 +411,53 @@ impl BodyBuilder {
...
+            condition,
+            |body| {
+                body.push_nothing();
+            },
+            else_builder,
+            responsible,
+        )
+    }
+    pub fn push_is_bool(&mut self, hir_id: &hir::Id, value: Id, responsible: Id) -> Id {
+        let is_condition_true = self.push_equals_value(value, true, responsible);
```
JonasWanke: This should be called something like `is_value_true` to match the generalized HIR ID below

JonasWanke (changes requested):

compiler/frontend/src/mir/body.rs
fpottbaecker: Should I also apply this to the rest of this class (I was matching the style of other methods in this class)?

compiler/frontend/src/hir_to_mir.rs
fpottbaecker: Moved to further up in this file

compiler/frontend/src/mir/body.rs
fpottbaecker: Maybe later?

compiler/frontend/src/string_to_rcst/expression.rs
JonasWanke: If you completely addressed a review comment and didn't write an answer, you can directly resolve the conversation yourself

compiler/frontend/src/mir/body.rs
JonasWanke: I think we should eventually use it where applicable, but that can be done over time. I think the feature is relatively recent, so we only used it for new code and didn't update everything yet

MarcelGarus (approved):

PR #841, opened 2023-12-14
Deduplicate deterministic expressions

MarcelGarus:
### Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

PR #871, opened 2024-01-11
Trace tail calls

JonasWanke:
**Depends on #841!**

### Checklist


compiler/frontend/src/mir_optimize/tail_calls.rs
```diff
@@ -0,0 +1,75 @@
...
+fn simplify_body(body: &mut Body) {
+    let [.., (call_id, Expression::Call { .. }), (
+        _,
+        Expression::TraceCallEnds {
+            return_value: end_return_value,
+        },
+    ), (_, Expression::Reference(reference_target))] = &mut body.expressions[..]
+    else {
+        return;
+    };
```
MarcelGarus, line 27: Does this work? Then we wouldn't need the if check below.
```suggestion
    let [.., (end_return_value, Expression::Call { .. }), (
        _,
        Expression::TraceCallEnds {
            return_value: end_return_value,
        },
    ), (_, Expression::Reference(end_return_value))] = &mut body.expressions[..]
    else {
        return;
    };
```

compiler/frontend/src/mir_optimize/tail_calls.rs
```diff
@@ -0,0 +1,75 @@
...
+                }
+
+                trace_call_starts = Some((
+                    index,
+                    *hir_call,
+                    *function,
+                    mem::take(arguments),
+                    *responsible,
+                ));
+                break;
```
MarcelGarus: Maybe this is clearer?

```suggestion
                if nesting == 0 {
                    trace_call_starts = Some((
                        index,
                        *hir_call,
                        *function,
                        mem::take(arguments),
                        *responsible,
                    ));
                    break;
                }
                nesting -= 1;
```

MarcelGarus (changes requested):

MarcelGarus (approved):

PR #887, opened 2024-01-18
Scopes higher order fn

### LINKED ISSUE #855: Add higher-order functions for starting and ending scopes
This pattern is all over the place in the AST to HIR compilation:

```rs
let reset_state = something.start.scope();
...
something.end_scope(reset_state);
```

Instead, we should use the `Drop` trait or, even better, have a higher-order function `create_scope`.

skyz1:
Closes: #855 

This PR adds a higher-order function to manipulate the context inside a scope. This means that scopes are no longer started and ended in different places.

### Checklist

### DIFF
--- a/compiler/frontend/src/ast_to_hir.rs
+++ b/compiler/frontend/src/ast_to_hir.rs
@@ -154,7 +154,7 @@ impl Context<'_> {
     }
 
     #[must_use]
-    fn with_scope<F, T>(&mut self, id_prefix: Option<hir::Id>, func: F) -> (Body, T)
+    fn with_scope<F, T>(&mut self, id_prefix: impl Into<Option<hir::Id>>, func: F) -> (Body, T)
     where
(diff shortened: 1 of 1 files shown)


compiler/frontend/src/ast_to_hir.rs
```diff
@@ -143,41 +143,42 @@ struct Context<'a> {
...
+        res
     }
-}
-struct NonTopLevelResetState(bool);
 
-impl Context<'_> {
     #[must_use]
-    fn start_scope(&mut self) -> ScopeResetState {
-        ScopeResetState {
+    fn with_scope<F, T>(&mut self, id_prefix: Option<hir::Id>, func: F) -> (Body, T)
```
JonasWanke: You can change `id_prefix` to accept an `impl Into<Option<hir::Id>>`, then callers can omit the `Some(…)` around that value

MarcelGarus (approved):

PR #929, opened 2024-02-03
Smaller HirId

### LINKED ISSUE #890: Make HirId smaller
The HIR ID is quite big and inefficient. It contains many `Vec`s, which point to heap storage and again to Strings. All of those pointers take up 8 bytes each.

If we would instead serialize the HIR ID's path to a string (such as `Example$fibonacci:2:11:28`), we'd save a lot of memory and pointer indirections. All operations on HIR IDs would become simple memcpy/memcmps.

skyz1:
Closes: #890

Saves the HirId path as a `String` instead of a `Vec`. This reduces the required storage space and increases performance.

The implementation makes the following assumptions:
- a named HirId-Key does not contain the characters `:` and `#`
- a named HirId-Key does not start with a numeric character
- a named HirId-Key does not end with a numeric character

### Checklist


compiler/frontend/src/module/module.rs
```diff
@@ -10,29 +10,59 @@ use std::{
...
 use tracing::{error, warn};
 
 #[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
-pub struct Module {
-    pub package: Package,
-    pub path: Vec<String>,
-    pub kind: ModuleKind,
+pub struct Module(Arc<InnerModule>);
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+struct InnerModule {
```
JonasWanke: The convention seems to be to use `Inner` as a suffix

```suggestion
struct ModuleInner {
```

compiler/frontend/src/module/module.rs
```diff
@@ -10,29 +10,59 @@ use std::{
...
+        &self.0.package
+    }
+    #[must_use]
+    pub fn path(&self) -> &Vec<String> {
+        &self.0.path
+    }
+    #[must_use]
+    pub fn kind(&self) -> ModuleKind {
+        self.0.kind
+    }
```
JonasWanke: These accessors should go below the “constructors” (`new(…)`, `from_path(…)`, …)

compiler/frontend/src/hir.rs
```diff
@@ -133,71 +134,125 @@ impl Body {
...
+pub struct IdPath(String);
 #[derive(Clone, Eq, From, Hash, Ord, PartialEq, PartialOrd)]
 pub enum IdKey {
     Named { name: String, disambiguator: usize },
     Positional(usize),
 }
+impl IdPath {
+    const fn empty() -> Self {
+        Self(String::new())
+    }
```
JonasWanke: You can instead derive `Default`

compiler/frontend/src/hir.rs
```diff
@@ -133,71 +134,125 @@ impl Body {
...
+
+    fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    fn push(&self, key: IdKey) -> Self {
```
JonasWanke: `push(…)` sounds like it's modifying this instance. Maybe rename it to `child(…)`? And please add `#[must_use]`

compiler/frontend/src/hir.rs
```diff
@@ -133,71 +134,125 @@ impl Body {
...
+            Self(format!("{}:{key}", self.0))
+        }
+    }
+
+    #[must_use]
+    pub fn last_as_string(&self) -> Option<&str> {
+        self.0.rfind(':').map(|i| &self.0[(i + ':'.len_utf8())..])
+    }
+
+    fn drop_last(&self) -> Option<Self> {
```
JonasWanke: `drop_last()` sounds like it's modifying this instance. Maybe rename it to `parent()`? And please add `#[must_use]`

compiler/frontend/src/hir.rs
```diff
@@ -133,71 +134,125 @@ impl Body {
...
+    }
+}
+impl From<IdKey> for IdPath {
+    fn from(value: IdKey) -> Self {
+        Self(format!("{value}"))
+    }
+}
+impl From<Vec<IdKey>> for IdPath {
+    fn from(value: Vec<IdKey>) -> Self {
+        Self(value.iter().map(|it| format!("{it}")).join(":"))
```
JonasWanke: `join(…)` formats the items automatically

```suggestion
        Self(value.iter().join(":"))
```

compiler/frontend/src/hir.rs
```diff
@@ -272,6 +332,7 @@ impl Debug for IdKey {
                 disambiguator,
             } => {
                 write!(f, "{name}")?;
+                // TODO(Optimization): Only include the `#` in Debug output and not when formatting for the HirId path
```
JonasWanke: Not sure if I understand this correctly. Do you want to drop all disambiguators in the `IdPath`? Because that would break things (we could no longer differentiate multiple assignments with the same name in the same scope)

compiler/frontend/src/hir.rs
```diff
@@ -207,46 +262,51 @@ impl Id {
...
-            .iter()
-            .map(|it| match it {
-                IdKey::Positional(index) => format!("<anonymous {index}>"),
-                IdKey::Named { name, .. } => name.to_string(),
+            .0
+            .split(':')
+            .map(|it| {
+                let first_char = it.chars().next().unwrap();
+                if first_char.is_numeric() {
+                    return format!("<anonymous {}>", it);
```
JonasWanke:
```suggestion
                    return format!("<anonymous {it}>");
```

compiler/frontend/src/hir.rs
```diff
@@ -133,71 +134,125 @@ impl Body {
...
+    fn push(&self, key: IdKey) -> Self {
+        if self.is_empty() {
+            Self::from(key)
+        } else {
+            Self(format!("{}:{key}", self.0))
+        }
+    }
+
+    #[must_use]
+    pub fn last_as_string(&self) -> Option<&str> {
```
JonasWanke:
```suggestion
    pub fn last_as_str(&self) -> Option<&str> {
```

compiler/frontend/src/hir.rs
```diff
@@ -133,71 +134,125 @@ impl Body {
...
+        self.0.rfind(':').map(|i| &self.0[(i + ':'.len_utf8())..])
```
JonasWanke: This might work as well (not tested):

```suggestion
        self.0.rsplit(':').next()
```

compiler/language_server/src/features_candy/analyzer/module_analyzer.rs
```diff
@@ -108,7 +108,8 @@ impl ModuleAnalyzer {
                     .unwrap();
                 let mut mir = (*mir).clone();
                 let mut static_panics = mir.static_panics();
-                static_panics.retain(|panic| panic.responsible.module == self.module);
+                static_panics
+                    .retain(|panic| -> bool { panic.responsible.module.clone() == self.module });
```
JonasWanke: The `module.clone()` doesn't seem to be necessary

```suggestion
                static_panics.retain(|panic| -> bool { panic.responsible.module == self.module });
```

compiler/vm/src/tracer/evaluated_values.rs
```diff
@@ -25,7 +25,7 @@ impl EvaluatedValuesTracer {
 impl Tracer for EvaluatedValuesTracer {
     fn value_evaluated(&mut self, heap: &mut Heap, expression: HirId, value: InlineObject) {
         let id = expression.get();
-        if id.module != self.module {
+        if id.module.clone() != self.module {
```
JonasWanke: The `module.clone()` doesn't seem to be necessary

```suggestion
        if id.module != self.module {
```

JonasWanke (changes requested):
Sorry for the delay!

Your last assumption (“a named HirId-Key does not end with a numeric character”) does not hold: A variable's name can end with a digit, and hence its HIR ID as well. But I think this is only slightly used by `id.function_name()` and that code correctly handles this case because then there's no `#`.

compiler/frontend/src/hir.rs
skyz1: You are right, that idea does not hold considering my last assumption was wrong.

JonasWanke (approved):
Looks good, thanks!

I'll wait for #926 to be merged first (hopefully today) so that we can see the results of this PR in the new benchmarking infrastructure

PR #1023, opened 2024-04-21
1016 leading whitespaces after indentation are not parsed correctly

### LINKED ISSUE #1016: Leading whitespaces after indentation are not parsed correctly
When parsing the string:
```
"
    foo
"
```
two of the white-spaces should be flagged as trailing white-space and two should remain with the string. Currently, both are marked as being trailing white-space.

The same error occurs when parsing
```
  foo
    bar
```
Here all four white-spaces in front of bar are marked as trailing, while only two of them should be.

styrix560:
Closes: #1016 

The problem was, that after parsing indentation whitespace-parsing was postponed. When the whitespaces consisted of two spaces, we first checked if you could parse indentation from the remaining input. Because of the two spaces you could, so more indentation was parsed. The solution was to not postpone whitespace-parsing and instead do it immediately after parsing indentation.

I am more or less sure the test cases look okay. If I were to choose, however, I would pick less. So, I would appreciate feedback.

### Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

### DIFF
--- a/compiler/frontend/src/string_to_rcst/whitespace.rs
+++ b/compiler/frontend/src/string_to_rcst/whitespace.rs
@@ -122,10 +122,7 @@ pub fn whitespaces_and_newlines(
     let mut new_input = input;
     let mut new_parts = vec![];
     let mut is_sufficiently_indented = true;
-    let mut current_indendation_level = 0;
-    println!("\ncalled with {input:?}, {indentation}\n");
     loop {
-        println!("indent: {current_indendation_level}/{indentation}");
         let new_input_from_iteration_start = new_input;
 
         if also_comments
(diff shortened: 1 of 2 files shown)


compiler/frontend/src/string_to_rcst/whitespace.rs
```diff
@@ -231,16 +231,18 @@ mod test {
...
         assert_rich_ir_snapshot!(
             whitespaces_and_newlines(" \n  foo", 0, true),
             @r###"
-        Remaining input: "  foo"
+        Remaining input: "foo"
         Parsed: Whitespace " "
         Newline "\n"
+        Whitespace "  "
         "###
         );
```
JonasWanke, line 247: Shouldn't the `Whitespace` here be wrapped in `CstError::TooMuchWhitespace` in the general case? (For parsed text, the output looks fine)

JonasWanke (changes requested):
Really sorry for the long delay! I somehow missed this PR

PR #1086, opened 2024-12-05
List methods

JonasWanke:
~**Depends on #1085!**~

### Checklist

### DIFF
--- a/packages_v5/example.candy
+++ b/packages_v5/example.candy
@@ -52,6 +52,27 @@ fun isAtLeast[T: Compare](left: T, right: T) Bool {
   }
 }
 enum Ordering { less, equal, greater }
+impl Ordering: Compare {
+  fun compareTo(self: Ordering, other: Ordering) Ordering {
+    switch self {
+      less => switch other {
+        less => Ordering.equal(),
+        equal => Ordering.less(),
+        greater => Ordering.less(),
+      },
+      equal => switch other {
+        less => Ordering.greater(),
+        equal => Ordering.equal(),
+        greater => Ordering.less(),
+      },
+      greater => switch other {
+        less => Ordering.greater(),
+        equal => Ordering.greater(),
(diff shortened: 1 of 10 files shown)


packages_v5/example.candy
```diff
@@ -368,11 +411,142 @@ fun replace[T](list: List[T], index: Int, item: T) List[T] {
...
+  list.allHelper(predicate, 0)
+}
+fun allHelper[T](list: List[T], predicate: (T) Bool, index: Int) Bool {
+  switch index.isLessThan(list.length()) {
+    true => switch predicate(list.get(index).unwrap()) {
+      true => list.allHelper(predicate, index.add(1)),
+      false => false,
+    },
+    false => true,
+  }
```
MarcelGarus: Maybe just do a get and switch on the option?

packages_v5/example.candy
```diff
@@ -368,11 +411,142 @@ fun replace[T](list: List[T], index: Int, item: T) List[T] {
...
+      .reduce((result: Text, item: Text) { "{result}, {item}" })
+      .unwrapOr("")
+    "[{items}]"
+  }
+}
+
+struct Pair[T0, T1] {
+  first: T0,
+  second: T1,
+}
```
MarcelGarus, line 542: In Martinaise, I use A and B as types and a and b as field names. Opinions?

(Although, in my experience, that pattern leads to compiler errors when you try to introduce a type named C as a compiler stage.)

MarcelGarus (approved):

MarcelGarus (approved):

packages_v5/example.candy
JonasWanke: I think I'll leave it as a generic `T` for now, but use different type parameter names if they correspond to something specific (like how we use a lifetime `'h` for referencing the HIR)

